PR #4, opened 2020-02-28
bug fix Readmi mnv package to mvn package

Melnyk9506:
bug fixed in Readmee

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -46,7 +46,7 @@ When starting the server, you need to initialize the database on the local machi
 In directory server run 
 
 ```
-mnv package
+mvn package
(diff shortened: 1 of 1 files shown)


vitaliimak (approved):

VanderbiGit (approved):
ok

r-yevchuk (approved):

narekacy:
1. Do we need inner .gitignore file inside **/server** package if we have it already in root?
2. Do we really need .mvn/ folder and files **mvnw.cmd**, **mvnw** in our repository in server path? As I know they are not required for building project.

Could you delete them from git observation?

narekacy:
Also could you delete feature branch after it was merged in dev? The previous one was not deleted

PR #10, opened 2020-03-07
Feature 7 add restaurant page

vitaliimak:
Don't pay attention to styles, html, etc. They don't work correctly.
Could you look did I add axios library well?


r-yevchuk (dismissed):

client/src/components/restaurant/restaurant.js
```diff
@@ -0,0 +1,34 @@
...
+    Api.getAll();
+  }
+
+  render() {
+    return (
+      <div>
+        <h2>Title</h2>
+        <Tabs defaultActiveKey="about">
+          <Tab eventKey="about" title="About">
+            <About />
```
VanderbiGit: We use tab?

client/src/services/api.js
```diff
@@ -0,0 +1,34 @@
...
+
+class Api {
+  constructor() {
+    this.apiUrl = '/data.json';
+    // 'http://localhost:8080/';
+  }
+
+  getAll() {
+    return axios.get(this.apiUrl)
+      .then((response) => {
```
VanderbiGit, line 10: axios work

VanderbiGit (commented):
you need connect to server

client/src/components/restaurant/restaurant.js
vitaliimak: Yes. According to design_decisions and mock-ups channels

vitaliimak:
> you need connect to server

I think, it's because of

> quote from slack
> Я спробував додати **фейк** бекенд

server/src/main/java/com/lunchtime/service/RestaurantService.java
```diff
@@ -0,0 +1,61 @@
...
+                restaurant.setEmail(newRestaurant.getEmail());
+                restaurant.setTextAddress(newRestaurant.getTextAddress());
+                restaurant.setWebsite(newRestaurant.getWebsite());
+                restaurant.setDescription(newRestaurant.getDescription());
+                restaurant.setWorkingTime(newRestaurant.getWorkingTime());
+                restaurant.setMenuId(newRestaurant.getMenuId());
+                restaurant.setOwnerId(newRestaurant.getOwnerId());
+                restaurant.setTables(newRestaurant.getTables());
+                restaurant.setLongitude(newRestaurant.getLongitude());
+                restaurant.setLatitude(newRestaurant.getLatitude());
```
vitaliimak, line 45: It doesn't look very good. Can someone know a better solution?

vitaliimak:
I think server part is ready to review.

vitaliimak:
Frontend part is ready to review too

client/src/components/restaurant/restaurant.js
```diff
@@ -0,0 +1,65 @@
...
+          <Tab eventKey="about" title="About">
+            <About restaurant={restaurant} isFetching={isFetching} />
+          </Tab>
+          <Tab eventKey="menu" title="Menu">
+            <h3>Menu</h3>
+          </Tab>
+          <Tab eventKey="events" title="Events">
+            <h3>Events</h3>
+          </Tab>
+          <Tab eventKey="feedback" title="Feedbac">
```
narekacy: "title" should be "**Feedback**", a little definition mistake

client/src/components/restaurant/restaurant.js
```diff
@@ -0,0 +1,65 @@
...
+          // eslint-disable-next-line no-console
+          console.error(response);
+          return;
+        }
+        this.setState({
+          restaurant: response.data,
+          isFetching: true,
+        });
+      });
+  }
```
narekacy: I think it is better to try to `setState()` before checking if response has got any error. I would switch lines 25 and 30 placing. 

Of course, we should check first `if(!response.error)` (or if we get exact response that is needed for `setState` method) and only after this `else`.

client/src/components/restaurant/restaurant.js
narekacy: I guess we need something like **Tab**-"wrapper" (or even **RestaurantTab**) component with props due to React.js. It will also help our code to be easier.

client/src/components/restaurant/restaurant.js
vitaliimak: I set `isFetching = false` in constructor so nested component won't break. If request returns error than `response.data = undefined` and we don't need to update state before if statement

server/src/main/java/com/lunchtime/controllers/RestaurantResource.java
```diff
@@ -0,0 +1,97 @@
...
+import org.springframework.web.bind.annotation.*;
+
+import javax.validation.Valid;
+
+@RestController
+@RequestMapping("/api/restaurants")
+public class RestaurantResource {
+
+    @Autowired
+    private RestaurantService restaurantService;
```
narekacy: Since Spring 4.3 it is recommended to avoid using `@Autowired` annotation but use **final** field (`private final RestaurantService restaurantService` in our case) and injecting by constructor.

```
public RestaurantResource (RestaurantService restaurantService){
        this.restaurantService= restaurantService;
    }
```

client/src/components/restaurant/restaurant.js
narekacy: I like your way. It seems good.

On the same time I guess we will have more successful

narekacy: I like your way. It looks good.

At the same time I guess we'll have more successful responses rather than failed. That's why I would prefer to check ones but not twice. 

**EDITED**:
to check what we will use for our purpose

server/src/main/java/com/lunchtime/controllers/RestaurantResource.java
```diff
@@ -0,0 +1,97 @@
...
+        if (restaurant.getId() != null) {
+            return ResponseEntity.badRequest()
+                .build();
+        }
+
+        Restaurant result = restaurantService.save(restaurant);
+        return ResponseEntity.created(new URI("/api/restaurants"))
+            .body(result);
+    }
+
```
narekacy, line 44: I think you meant `if (restaurant.getId() == null)` in line 33

server/src/main/java/com/lunchtime/service/RestaurantService.java
narekacy: For sure, Hibernate has prepared something useful for us. [**Check this**](https://stackoverflow.com/questions/2398981/effective-entity-update-in-hibernate), please

Andrii35 (commented):
By creating of entities we need to consider the relationships between them.  You have in Restaurant entity two foreign keys "menu_id" and "owner_id".  I found information, that we need to use annotations "@OneToOne", "@OneToMany" or "@ManyToMany". How to do it you can read here https://www.javaworld.com/article/2077819/understanding-jpa-part-2-relationships-the-jpa-way.html or  https://easyjava.ru/data/jpa/jpa-i-svyazi-mezhdu-obektami/

vitaliimak:
> By creating of entities we need to consider the relationships between them...

You are right, but in this task I do only restaurant part, I can't add these relations because tables "user", "menu", etc don't exist. Person who creates these entities should do it.

So I added temporary fake fields

server/src/main/java/com/lunchtime/service/RestaurantService.java
vitaliimak: Thanks for answer. But `dynamicUpdate=true` doesn't help in this case. 
This code looks bad because developer must look after 2 places in codebase. When entity Restaurant is updated then method update(restaurant) must be updated too. 
I can't save newRestaraunt object directly, because it comes from api request and can contain fields like createdAt with wrong data (or other wrong fields).

client/src/components/restaurant/restaurant.js
vitaliimak: Sorry, I'm not sure that understand you

client/src/components/restaurant/restaurant.js
vitaliimak: In this case
```
then((response) => {
        this.setState({
          restaurant: response.data,
          isFetching: true,
        });
        if (response.error) {
          // eslint-disable-next-line no-console
          console.error(response);
          this.setState({
            restaurant: {},
            isFetching: false,
          });
          // return;
        }
      });
```
I have this result if error happens
image

I use Error-First principle to handle errors http://fredkschott.com/post/2014/03/understanding-error-first-callbacks-in-node-js/
It doesn't affected user experience because if statement works very fast

But anyway, could you show code example with your solution?

server/src/main/java/com/lunchtime/controllers/RestaurantResource.java
vitaliimak: Thanks, fixed

server/src/main/java/com/lunchtime/controllers/RestaurantResource.java
vitaliimak: No if new Restaurant has id (restaurant.getId() != null) then user must use update request, because client can't know id in new Restaurant entity. Only database creates id in every new entities

server/src/main/java/com/lunchtime/service/RestaurantService.java
narekacy: I think we should hold this place for future and return here. There must be more flexible way to update our entity. Now we can't imagine Hibernate's power, but we will consider to improve it later. It's ok.

client/src/components/restaurant/restaurant.js
narekacy:
```
      .then((response) => {
        if (!response.error) {
            this.setState({
                restaurant: response.data,
                isFetching: true,
            });
            return;
        } else {
            console.error(response);
        }
      });
```

server/src/main/java/com/lunchtime/controllers/RestaurantResource.java
narekacy: Got it. Thanks!

client/src/components/restaurant/restaurant.js
vitaliimak: I try to avoid negative conditionals https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29#understandability-tips

It makes code less readable

VanderbiGit:
Screenshot from 2020-03-10 11-41-55

I have some error. In your computer work well?

vitaliimak:
> I have some error. In your computer work well?

It's because of very magic annotations `@Setter and @Getter` in Restaurant class. Those annotations generate getters and setters in runtime and IDEA doesn't know about those

Code compiles well. But you can fix those errors in ide. Just install plugin https://plugins.jetbrains.com/plugin/6317-lombok and update settings https://stackoverflow.com/questions/17729384/lombok-added-but-getters-and-setters-not-recognized-in-intellij-idea

VanderbiGit:
What about frontend?
Screenshot from 2020-03-10 11-53-10

vitaliimak:
> What about frontend?

I hardcoded restaurantId  from my databse. This id should send dynamically from restaurant list. It will do other developer. As you can see, even when id doesn't exist application doesn't crash and does show `Loading...` message.

You can create new Restaurant entity in your database (from Postman) and update restaurantId in restaurant-list.js for testing

server/src/main/java/com/lunchtime/models/Restaurant.java
```diff
@@ -0,0 +1,89 @@
...
+    @Email
+    @Column(name = "email")
+    private String email;
+
+    @NotBlank
+    @Column(name = "text_address")
+    private String textAddress;
+
+    @Column(name = "website")
+    private String website;
```
narekacy, line 41: There is `@URL` annotation for field's validation as a website. But we should take care about its regex: 

_Per default the constraint validator for this constraint uses the java.net.URL constructor to validate the string. This means that a matching protocol handler needs to be available. Handlers for the following protocols are guaranteed to exist within a default JVM - http, https, ftp, file, and jar. See also the Javadoc for URL._  -> [link](https://softwareengineering.stackexchange.com/questions/356540/how-to-represent-and-validate-website-urls-for-a-jpa-entity)

server/src/main/java/com/lunchtime/models/Restaurant.java
```diff
@@ -0,0 +1,89 @@
...
+    private Long createdBy;
+
+    @Column(name = "modify_at")
+    @UpdateTimestamp
+    private Instant modifyAt;
+
+    @Column(name = "modify_by")
+    private Long modifyBy;
+
+    public Boolean getIsDeleted() {
```
narekacy: Due to JavaBean Specifications (**8.3.2 Boolean properties**), we should use:
public boolean (or Boolean) **is**Deleted<PropertyName>() instead of **getIs**Deleted<PropertyName>()

Also we can use a ternary operator for code's readability

VanderbiGit (dismissed):

Andrii35 (dismissed):

r-yevchuk (dismissed):

server/src/main/java/com/lunchtime/models/Restaurant.java
vitaliimak: Thanks, very useful annotation! But by default it forbids urls like `somelink.domain.com` (without protocol). It can be customizable by regex, as I understand.
As our business process requires be moderator who check restaurant entity before approve. I think this annotation will bring more complexity then advantage. E.g. if this field must be valid there is must to be validation in input in frontend
I also didn't add annotation `@NotBlank` as restaurant may not have website

server/src/main/java/com/lunchtime/models/Restaurant.java
vitaliimak: Unfortunately I didn't find this rule, but I think you're right

server/src/main/java/com/lunchtime/models/Restaurant.java
narekacy: I think we should embody possible complexity to advantage :) Of course we should add regex so that user can see that his web-URL is not valid, but not moderator.

i.e.: coffee-cafe.com (where first '**c'** is not latin one: this can happen when user starts to type and doesn't switch keyboard language. So, when he notices this he backspace his not-latin letters up to 'c', thinking that 'c' on cyrillic and 'c' on latin is the same symbol :D

server/src/main/java/com/lunchtime/models/Restaurant.java
narekacy: Thank you for reliance! :)

**[Here ](https://download.oracle.com/otndocs/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/)** you can download it and see **8.3.2** paragraph. I also attached an image from this file, supporting this recommendation.

t-bp

server/src/main/java/com/lunchtime/models/Restaurant.java
vitaliimak: Regular expressions are very powerful feature and very dangerous and complicated. I think we need to review our requirements according to this conversation and will decide what way is best

There are few references about email:
https://habr.com/en/post/175375/
https://habr.com/en/post/274985/

There is RFC about website naming (I didn't read):
https://tools.ietf.org/html/rfc819

narekacy: I could be wrong. I am going to maintain your opinion. It will be the best decision to review our requirements.

When reviewing, let us remember that this "micro" feature can add point to our developer skills but only if we use smart solution in short time:)

Anyway, may be this "micro" feature is not worth this conversation?)

Sorry in advance :)

server/src/main/java/com/lunchtime/models/Restaurant.java
vitaliimak: I am completely confused)
Should I rename property `private Boolean isDeleted` to `private Boolean deleted`?
Because pattern `get<PropertyName>` where `PropertyName` is `isDeleted` was realized in previous code iteration

narekacy: If I understand you correctly (line 19), I would keep property `private Boolean isDeleted` and rename`getIsDeleted()` to `isDeleted()`. And we will have

```
private Boolean isDeleted
public Boolean isDeleted()
```

vitaliimak: Fixed

client/src/components/restaurant/restaurant.js
vitaliimak: @narekacy Could you explain it better?

narekacy: Check this [video tutorial](https://www.youtube.com/watch?v=-mDs48HB3II&list=PLcvhF2Wqh7DNVy1OCUpG3i5lyxyBWhGZ8&index=16), please. You may set 1.5 or even 2.0 playback speed, this guy explains it in excellent way. 

When you finish watching it, remind yourself that you also can use an array to pass in it your component as props with `map()` magic.

VanderbiGit (approved):

Andrii35 (approved):

r-yevchuk (approved):

PR #11, opened 2020-03-08
Feature 11 events page backend yevchuk

r-yevchuk:


server/src/main/java/com/lunchtime/controllers/EventController.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+
+import java.sql.Date;
+import java.util.List;
+
+@RestController
+@RequestMapping("/events")
```
vitaliimak: Should we add "/api" prefix in every endpoints? As it is common practice in REST services (e.g. github, google, etc)

server/src/main/java/com/lunchtime/controllers/EventController.java
```diff
@@ -0,0 +1,55 @@
...
+    public ResponseEntity<?> getByDateBetween(@PathVariable("startDate") Date startDate,
+                                              @PathVariable("endDate") Date endDate) {
+        List<Event> result = eventService.findByDateBetween(startDate, endDate);
+        return new ResponseEntity<>(result, HttpStatus.OK);
+    }
+
+    @PostMapping
+    public ResponseEntity<?> addOrUpdateEvent(@RequestBody Event event) {
+        eventService.saveOrUpdateEvent(event);
+        return new ResponseEntity<>("Event added successfully", HttpStatus.OK);
```
vitaliimak: POST method should  return created entity, it helps frontend developer to understand that server created entity correctly. Also response status code should be 201 - CREATED.

server/src/main/java/com/lunchtime/controllers/EventController.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @GetMapping("date/{startDate}/{endDate}")
+    public ResponseEntity<?> getByDateBetween(@PathVariable("startDate") Date startDate,
+                                              @PathVariable("endDate") Date endDate) {
+        List<Event> result = eventService.findByDateBetween(startDate, endDate);
+        return new ResponseEntity<>(result, HttpStatus.OK);
+    }
+
+    @PostMapping
+    public ResponseEntity<?> addOrUpdateEvent(@RequestBody Event event) {
```
vitaliimak: image

You should add minimal validation and return Bed Request if user send stupid data. 
Don't believe user!

server/src/main/java/com/lunchtime/controllers/EventController.java
```diff
@@ -0,0 +1,55 @@
...
+        eventService.saveOrUpdateEvent(event);
+        return new ResponseEntity<>("Event added successfully", HttpStatus.OK);
+    }
+
+    @DeleteMapping
+    public void deleteEvent(@RequestParam("id") long id) {
```
vitaliimak: You should use `@DeleteMapping("{id}")` as it is simpler and https://www.restapitutorial.com/lessons/httpmethods.html

server/src/main/java/com/lunchtime/controllers/EventController.java
```diff
@@ -0,0 +1,55 @@
+package com.lunchtime.controllers;
```
vitaliimak, line 1: There aren't methods `getOneById(id) and update(id)`

server/src/main/java/com/lunchtime/controllers/EventController.java
```diff
@@ -0,0 +1,55 @@
...
+    EventService eventService;
+
+    @GetMapping
+    public ResponseEntity<?> getAll() {
+        List<Event> result = eventService.findAll();
+        return new ResponseEntity<>(result, HttpStatus.OK);
+    }
+
+    @GetMapping("/{category}")
+    public ResponseEntity<?> getByCategory(@PathVariable("category") String category) {
```
vitaliimak: For now it isn't so important. But as you implemented this method, I think you should use enum instead of String. Because in current implementation user can sent any request e.g. `GET localhost:8080/events/date` and spring app makes useless request to database

server/src/main/java/com/lunchtime/entity/Event.java
```diff
@@ -0,0 +1,35 @@
...
+import lombok.Setter;
+import lombok.ToString;
+
+import javax.persistence.*;
+import java.net.URI;
+import java.sql.Date;
+
+@ToString
+@Setter
+@Getter
```
vitaliimak, line 16: Incredible! Thanks for this magic!

server/src/main/java/com/lunchtime/entity/Event.java
```diff
@@ -0,0 +1,35 @@
...
+    private String eventName;
+    private String category;
+    private String description;
+
+    /**
+     * An empty constructor is needed to create a new instance via
+     * reflection by persistence framework.
+     */
+    public Event() {
+    }
```
vitaliimak: I did similar class, and code works without empty constructor. Is it really need? As Java class has empty constructor by default.

server/src/main/java/com/lunchtime/service/EventServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    public List<Event> findByDateBetween(Date startDate, Date endDate) {
+        return eventRepository.findByDateBetween(startDate, endDate);
+    }
+
+    public void saveOrUpdateEvent(Event event) {
+        eventRepository.save(event);
+    }
+
+    public void deleteEvent(long id) {
+        eventRepository.deleteById(id);
```
vitaliimak, line 42: If user send wrong id (e.g. 999999999999) server throw Error 500. It's cannot be allowed!

server/src/main/java/com/lunchtime/LunchTimeApplication.java
```diff
@@ -2,12 +2,28 @@
...
+@SpringBootApplication(
+    exclude = {SecurityAutoConfiguration.class, SecurityAutoConfiguration.class})
 public class LunchTimeApplication {
 
     public static void main(String[] args) {
         SpringApplication.run(LunchTimeApplication.class, args);
     }
 
+    /*Set the CORS configuration, more about CORS:
+      https://www.tutorialspoint.com/spring_boot/spring_boot_cors_support.htm */
```
vitaliimak: You don't need this comment. Your code speaks for himself.
https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29#comments-rules

vitaliimak (changes requested):
Great work! But there are some issues.
Also what do you think if we disable all checkstyle rules which connected to imports? I mean 
```
<module name="AvoidStarImport"/>
<module name="CustomImportOrder">
      <property name="sortImportsInGroupAlphabetically" value="true"/>
      <property name="separateLineBetweenGroups" value="true"/>
      <property name="customImportOrderRules" value="STATIC###THIRD_PARTY_PACKAGE"/>
      <property name="tokens" value="IMPORT, STATIC_IMPORT, PACKAGE_DEF"/>
</module>
```
They are really annoying

server/src/main/java/com/lunchtime/controllers/EventController.java
```diff
@@ -0,0 +1,55 @@
...
+        return new ResponseEntity<>(result, HttpStatus.OK);
+    }
+
+    @GetMapping("date/{date}")
+    public ResponseEntity<?> getByDate(@PathVariable("date") Date date) {
+        List<Event> result = eventService.findByDate(date);
+        return new ResponseEntity<>(result, HttpStatus.OK);
+    }
+
+    @GetMapping("date/{startDate}/{endDate}")
```
vitaliimak: You can use more readable format in this case `/date?from=date&to=date`

VanderbiGit:
round off  back and front in one pull request

server/src/main/java/com/lunchtime/entity/Event.java
```diff
@@ -0,0 +1,35 @@
...
+@Table(name = "events")
+public class Event {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    private long id;
+    private long restaurantId;
+    private Date date;
+    private URI eventImage;
+    private String eventName;
```
vitaliimak: You don't need use prefix event in naming as it is Event entity

PR #19, opened 2020-03-12
Feature 11 events page yevchuk

r-yevchuk:


narekacy (commented):
[test]

client/src/style/m-button.css
```diff
@@ -0,0 +1,22 @@
...
+.m-button:focus {
+  background: none;
+  border-color: #fac564;
+  box-shadow: none;
+}
+
+.m-button:active{
+  background: #fac564 !important;
+  border-color: #fac564 !important;
+  box-shadow: none !important;
```
narekacy, line 21: We should avoid `!important`

server/src/main/java/com/lunchtime/controllers/EventController.java
```diff
@@ -0,0 +1,33 @@
...
+import javax.validation.Valid;
+import java.util.Date;
+import java.util.List;
+
+@RestController
+@RequestMapping("/api/events")
+public class EventController {
+
+    @Autowired
+    EventService eventService;
```
narekacy: Use constructor injecting instead of `@Autowired` annotation:

```
private final EventService eventService;

public EventController(EventService eventService) {
         this.eventService = eventService;
}
```

The same as line 15 [here](https://github.com/Social-projects-Rivne/Lunch_Time/pull/19/files?file-filters%5B%5D=.css&file-filters%5B%5D=.java&file-filters%5B%5D=.js&file-filters%5B%5D=.json#diff-0607fa89e7d13943be76841f7d4b65c9)

server/src/main/java/com/lunchtime/controllers/EventController.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @GetMapping
+    public ResponseEntity<?> getAll() {
+        List<Event> result = eventService.findAll(new Date());
+        return new ResponseEntity<>(result, HttpStatus.OK);
+    }
+
+    @PostMapping
+    public ResponseEntity<?> addEvent(@RequestBody @Valid Event event) {
```
Andrii35: Should't be the anotation  @Valid for @RequestBody

server/src/main/java/com/lunchtime/entity/Event.java
```diff
@@ -0,0 +1,49 @@
...
+    @NotBlank
+    private String category;
+
+    @NotBlank
+    @Size(min=6, max=999)
+    private String description;
+
+    @NotNull
+    private Boolean isActive;
+}
```
Andrii35: Anatotion @Column is not obligatory, but for better understanding is it better, if we use this anotation, we can describe here column parameters too.

Andrii35: You have created anothe package "entity" for your model "event". We have package models, where we create all our models(entities)

server/src/main/java/com/lunchtime/repository/EventRepository.java
```diff
@@ -0,0 +1,13 @@
...
+import com.lunchtime.entity.Event;
+import org.springframework.data.repository.CrudRepository;
+
+import java.util.Date;
+import java.util.List;
+
+
+public interface EventRepository extends CrudRepository<Event, Long> {
+
+    List<Event> findByDateGreaterThanAndIsActiveTrueOrderByDateAsc(Date date);
```
Andrii35, line 16: The name of method is to long. Method should make one handle. If they make more that one, you schould make another method. KISS principle.

client/src/style/m-button.css
r-yevchuk: I know, but can't do this without this tag

server/src/main/java/com/lunchtime/controllers/EventController.java
r-yevchuk: I do not agree. When I delete annotation and add a new object with wrong parameters, i get server error.

server/src/main/java/com/lunchtime/repository/EventRepository.java
r-yevchuk: What is an alternative?

vitaliimak: You can use few simple methods and connect them

server/src/main/java/com/lunchtime/controllers/EventController.java
Andrii35: I mean, you should use in this case (@Valid @RequestBody Event eventt). Also @Valid must stay for @RequestBody

server/src/main/java/com/lunchtime/repository/EventRepository.java
vitaliimak: Nice magic! I didn't know about Query by Example.
In this case your code looks well, but I think you should left method `findAll` and add method `findAllSorted` or `findAllActive`. Because `findAll` can be misleading in this case

client/src/pages/events.js
```diff
@@ -0,0 +1,71 @@
...
+          events: response.data,
+        });
+      });
+  }
+
+  render() {
+    const { events } = this.state;
+
+    return (
+      <Container fluid className="page-container p-0">
```
vitaliimak, line 46: image

What will happen when database is empty or server return error, or network brakes down? I think, you don't need to change this PR, but you can show some `Loading message` or spinner in this case
https://react-bootstrap.github.io/components/spinners/

server/src/main/java/com/lunchtime/models/Event.java
```diff
@@ -0,0 +1,56 @@
...
+    @Temporal(TemporalType.TIMESTAMP)
+    @Column(name = "date")
+    private Date date;
+
+    @NotNull
+    @Column(name = "image")
+    private URI image;
+
+    @NotBlank
+    @Size(min=6, max=255)
```
vitaliimak: You should disable rule `WhitespaceAround` in checkstyle or update your code as it destroys building process
```
lunch-time-server | [ERROR] /app/src/main/java/com/lunchtime/models/Event.java:40:14: WhitespaceAround: '=' is not followed by whitespace. Empty blocks may only be represented as {} when not part of a multi-block statement (4.1.3) [WhitespaceAround]
lunch-time-server | [ERROR] ...
```

client/package.json
```diff
@@ -8,14 +8,19 @@
     "@testing-library/user-event": "^7.1.2",
     "axios": "0.19.2",
     "bootstrap": "4.4.1",
+    "lodash": "4.17.15",
+    "moment": "^2.24.0",
+    "node-uuid": "^1.4.8",
```
vitaliimak: Where do you use this library?

client/src/components/my-dropdown.js
```diff
@@ -0,0 +1,34 @@
...
+  render() {
+    const { id, name, items } = this.props;
+    return (
+      <Dropdown className="mr-3">
+        <Dropdown.Toggle id={id} className="m-button">{name}</Dropdown.Toggle>
+        <Dropdown.Menu>
+          {items.map((item) => (
+            <Dropdown.Item
+              className="m-dropdown-item"
+              key={uuid.v4()}
```
vitaliimak: Why do you use this additional library, but not item.id?

vitaliimak: If it is array with string and not objects with property id, why don't you use array index? As I understand you don't change this array, so nothing bad will happen

> Let me explain, a key is the only thing React uses to identify DOM elements. What happens if you push an item to the list or remove something in the middle? If the key is same as before React assumes that the DOM element represents the same component as before. But that is no longer true

server/src/main/java/com/lunchtime/controllers/EventController.java
vitaliimak: @Andrii35 could you explain, why must this annotation order be exactly this?

Andrii35: I read many examles about this annotations. It is nothing is told what the difference is where the annotation stands. But in only one case did I see the location as in Roman. For better readability of the code I would put @Valid for @RequestBody.

vitaliimak (approved):

Andrii35 (approved):

narekacy (approved):

PR #20, opened 2020-03-13
was created table entities ( user ,role ,status, baseentytyts)  everi…

Melnyk9506:
User, roles, status

### DIFF
--- a/server/src/main/java/com/lunchtime/models/BaseEntity.java
+++ b/server/src/main/java/com/lunchtime/models/BaseEntity.java
@@ -1,0 +1,37 @@
+package com.lunchtime.models;
+
+import lombok.Data;
+import org.springframework.data.annotation.CreatedDate;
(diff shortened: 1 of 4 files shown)


vitaliimak (changes requested):
Unfortunately without Controller class and without POST method in Controller class it isn't possible to review this PR.
Also you should follow to our code style

server/src/main/java/com/lunchtime/models/BaseEntity.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Date;
+
+
+
+@MappedSuperclass
+@Data
+public class BaseEntity {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
```
Andrii35, line 23: You should change GenerationType.IDENTITY to GenerationType.AUTO becouse IDENTITY used with MySQL DB.

PR #22, opened 2020-03-17
Feature 18 add personal info page makarchenkov

vitaliimak:


Melnyk9506 (commented):
you wrote a restaurant service only there is already an interface with that name
and i have a proposition off the chechstyle 
and isDeleted in Person Entity i think we need to do the enum class Status and write different statuses not even deleted

vitaliimak:
> you wrote a restaurant service only there is already an interface with that name
> and i have a proposition off the chechstyle
> and isDeleted in Person Entity i think we need to do the enum class Status and write different statuses not even deleted

I'm not sure that I understand you.
Do you propose disable checkstile linter? If yes, you can tell it in daily meeting, it doesn't connect to my task.
In my task I only updated Restaurant service. I made connection between new entity Person and old entity Restaurant and created Restaurant service interface. And I've changed file name according to project naming convention (RestaurantResource -> RestaurantController).

`isDeleted` property shows that entity is marked as not exist to backend application. If you want to add some business statuses, like `isOpen, isClosed` and so on, I think, you should add new property `status`

Andrii35 (commented):
If we make diferents between Services and Services Implementation, we should make also two diferent packages. Services(for interfaces) and Implementation(for classes).
I have checked  server part with postman. (GET and POST requests works successfully).
Client part works and run successfully to.

vitaliimak:
> If we make diferents between Services and Services Implementation, we should make also two diferent packages. Services(for interfaces) and Implementation(for classes).

All right. I'll do it today later

vitaliimak:
@Andrii35 I've created implementation package. Could you review if it is correct?

Andrii35 (commented):
Ok, package look's good, but I would call classes like "class RestaurantServiceImplement implements RestaurantService{}. You have deleted this names.

vitaliimak:
> Ok, package look's good, but I would call classes like "class RestaurantServiceImplement implements RestaurantService{}. You have deleted this names.

Done

client/src/components/profile.js
```diff
@@ -0,0 +1,107 @@
...
+          </h1>
+        </Container>
+        <Container className="card-body pl-5 pr-5">
+          <Row>
+            <Col md="3" className="profile-menu">
+              <ListGroup>
+                {this.menuItems.map((menuItem, index) => {
+                  return (
+                    // eslint-disable-next-line react/no-array-index-key
+                    <Link to={menuItem.path} key={index}>
```
narekacy, line 71: What 'index' refers to here?

vitaliimak: In this case index is index of array element https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map
I use it because of that `menuItems` array is static.

https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318
> Update: Exception from the rule
> 
> Many people asked if they always, always have to generate ids. Others have suggested use cases when using the index as a key seems justifiable.
> 
> It is true that sometimes generating new ids is redundant and may be avoided. For example translation of license terms or list of contributors.
> 
> To help you decide, I put together three conditions which these examples have in common:
> 
>     the list and items are static–they are not computed and do not change;
>     the items in the list have no ids;
>     the list is never reordered or filtered.
> 
> When all of them are met, you may safely use the index as a key.

narekacy: Nice. Thank you!

client/src/components/profile/info.js
```diff
@@ -0,0 +1,56 @@
...
+              <p>
+                Email:
+                {' '}
+                { user.email }
+              </p>
+              <p>
+                Phone:
+                {' '}
+                { user.phoneNumber }
+              </p>
```
narekacy, line 31: I think it is better create 'UserInfo' component and reuse lines 17-35 there.

Andrii35 (dismissed):
Good

server/src/main/java/com/lunchtime/models/Person.java
```diff
@@ -0,0 +1,71 @@
...
+    private Instant modifyAt;
+
+    @Column(name = "modify_by")
+    private Long modifyBy;
+
+    public Boolean isDeleted() {
+        if (isDeleted == null) {
+            return false;
+        }
+        return isDeleted;
```
narekacy, line 65: Use ternary operator here

client/src/components/profile/info.js
vitaliimak: It is userinfo component (info.js) and you can reuse it everywhere, you only need to send object with properties `email, name, etc` into this component.
Splitting this component more makes code and project more complicated and doesn't bring any advantages
And now this code works and does their function - show information about User

Then and only then when we need to split this component we will do it https://martinfowler.com/bliki/Yagni.html

server/src/main/java/com/lunchtime/models/Person.java
vitaliimak: No I don't like ternary operator, in my opinion it makes code less readable. I try to avoid it. But you can use it.

client/src/components/profile/info.js
narekacy: Sorry, I was mistaked - we can use this component `as UserInfo` component anywhere. 

Anyway, lines 18, 23 and 28 are hardcoded. It is better:
1. to define them in default prop values;
2. use `.map()` method to retrieve all incoming prop values.

vitaliimak: Sorry, I don't understand you. I did my task and why I need to think about "something anywhere"? Right now this component is used only in one place. And I can't see future (unfortunately) and I don't now if we will need to use this component in other place of project.

About creating menu lines from array. There are only 3 menu items (avatar has other structure). Creating additional abstraction layer brings more complexity than advantages in this case. If we will see (in the future) that we need more menu items we can change it in any time

narekacy: Ok

Melnyk9506 (dismissed):

server/src/main/java/com/lunchtime/implementation/RestaurantServiceImplement.java
```diff
@@ -0,0 +1,74 @@
...
+            .orElseGet(() -> {
+                return null;
+            });
+    }
+
+    public Restaurant delete(Long id) {
+        return findById(id)
+            .map(restaurant -> {
+                restaurant.setIsDeleted(true);
+                return save(restaurant);
```
narekacy, line 68: I think we should use `setisDeleted()` method with no args for safety. Also I'd rename it to `setDeleted()` and create another method `setAvailable()` or `setVisible()` or even `setNotDeleted()`, where we can "restore" entity from deleted ones if needed.

Also take care about `.map()` method. This code looks very cool with `.map()`, but it refers more to multiple elements and/or collections.

server/src/main/java/com/lunchtime/models/Restaurant.java
```diff
@@ -54,8 +51,10 @@
     @Column(name = "menu_id")
     private Long menuId;
 
-    @Column(name = "owner_id")
-    private Long ownerId;
+    @ManyToOne
+    @NotNull
+    @JoinColumn(name = "owner_id", nullable = false)
```
narekacy: We planned just with a little mistake our database. `owner_id` should be `user_id`.

server/src/main/java/com/lunchtime/implementation/RestaurantServiceImplement.java
narekacy: Or even better. We can use `setInactive()` and `setActive()` methods. In fact, we don't delete entities, but mark them as non-shown.

vitaliimak: I agree that we can change `setDelete()` method signature. But it must be delete, because user push button delete and he or she expects that entity will be deleted. Also there is common practice in this property naming.
`setInActive() setActive()` can be applied to `Person` who doesn't confirm email

vitaliimak: @narekacy 
I cant find information that using `.map()` in this case is a bad practice. Optional interface provide this method and repository can use it.
https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html
https://docs.spring.io/spring-data/commons/docs/current/api/org/springframework/data/repository/CrudRepository.html#findById-ID-
Could you explain why it is bad practice?

vitaliimak: As my task is "Show personal info" I created new task in trello to fix this issue https://trello.com/c/HB5kLggh/95-90-change-setdelete-method-signature

narekacy: > Also take care about `.map()` method. This code looks very cool with `.map()`, but it refers more to multiple elements and/or collections.

No no, I don't call it bad practiced, no :) I just wanted to express my opinion that `.map()` in Java is more about Stream API for collections, that's it :)

This code looks great and I like it for sure, a fortiory, it works!

vitaliimak: Thanks for explanation. I'm newbie in Steam API and I didn't find any mentions about `stream` in `java.util.Optional` class

server/src/main/java/com/lunchtime/models/Restaurant.java
vitaliimak: Fixed, but as we use `Person` entity (not `User` - Spring creates `user` table by default for system user) I changed column name to `person_id`

vitaliimak:
@Andrii35 @Melnyk9506 @narekacy I made few changes, so could you review PR again?

server/src/main/java/com/lunchtime/models/Restaurant.java
narekacy: Sure :)

narekacy (approved):
Nice!

Andrii35 (approved):

Melnyk9506 (approved):

PR #27, opened 2020-03-23
Feature 79 restaurant list voznyarskyy

VozNick:


server/src/main/java/com/lunchtime/models/RestaurantVisit.java
```diff
@@ -0,0 +1,27 @@
+package com.lunchtime.models;
```
vitaliimak, line 1: Why you create this model? We have restaurant model. A lot of users (persons) use our application, will you plane to create separate RestaurantVisit entity for every user?
Is your task need this functionality?

server/src/main/java/com/lunchtime/implementation/RestaurantServiceImplement.java
```diff
@@ -40,6 +40,36 @@ public Restaurant save(Restaurant restaurant) {
...
+    }
+
+    @Override
+    public Page<Restaurant> yourPastVisits(Pageable pageable) {
+        return null;
+    }
+
+    @Override
+    public Page<Restaurant> chooseType(Pageable pageable) {
+        return null;
```
vitaliimak, line 70: If this code is useless you shouldn't merge it in dev branch

client/src/components/info/restaurants.js
```diff
@@ -0,0 +1,47 @@
+const sortBy = ['Relevance', 'Distance', 'Your match'];
+const city = ['Вінницька область', 'Волинська область', 'Дніпропетровська область', 'Донецька область',
```
vitaliimak: Don't use non Latin letters in code. If you need localization, use special tools for it

vitaliimak (changes requested):

server/src/main/java/com/lunchtime/models/RestaurantVisit.java
VozNick: I mean it`s for sorting reason...
in restaurant list page, you can sort restaurants by your last visit. This class will keep dates of all rest. that you have visited...

vitaliimak: Many users use this app.
Bob visited Restaurant A in 2020-03-05
Maria visited Restaurant A in 2020-03-06
Bob visited Restaurant B in 2020-03-06
...

Your model doesn't connect to Persons, how you prevent collisions between Persons? How you save date 2020-03-05 (Bob) and 2020-03-06 (Maria)? Does this functionality need?

PR #31, opened 2020-03-26
Feature 91. Fix 'Reset button'. narek

narekacy:
This should work, but I have no idea why `dropdown-group.js` line 14 is underlined like undefined. 

Due to React.js we should use one state in order to manage our component. Unfortunately, this doesn't work / or I'm doing something wrong. 
By the way: lines 42 and 61 in `dropdown-group.js` are commented due to eslint rules. You'll need to use it while reviewing.

Check this code, and give me a feedback. please.
Thank you in adnvance.

### DIFF
--- a/client/src/components/dropdown-group.js
+++ b/client/src/components/dropdown-group.js
@@ -10,6 +10,7 @@ class DropdownGroup extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
+      // eslint-disable-next-line react/no-unused-state
       categories: [], // we have empty array, that pass to CategoryDropdown
     };
     this.onAddCategory = this.onAddCategory.bind(this);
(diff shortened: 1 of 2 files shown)


client/src/components/dropdown-group.js
```diff
@@ -23,6 +39,9 @@ class DropdownGroup extends React.Component {
           items={eventsType}
           homePath="events"
           path="events/categories/"
+          {/* eslint-disable-next-line no-undef */}
+          categories={categories} // here we pass our empty array (or updated later)
```
vitaliimak, line 43: You don't have variable `categories` in function render.
`const { categories } = this.state;`

narekacy: Thanks a lot

vitaliimak:
Your branch is called `feature_number` but it must be `fix_number` 
I think it isn't important right now, but you should more attentive

client/src/components/dropdown-group.js
narekacy: Hmm.. but we can't consider `categories` 'used' if we pass it as a prop? )

narekacy: It seems I answered my question :) It works. Thank you

narekacy:
Thank you

PR #34, opened 2020-03-27
Fix 91. 'Reset button'. narek

narekacy:
It should be in working condition. But it doesn't work.
You will also need lines 20-24 in `dropdown-group.js`.

### DIFF
--- a/client/src/components/dropdown-group.js
+++ b/client/src/components/dropdown-group.js
@@ -7,6 +7,23 @@ import MyDropdown from './my-dropdown';
 import CategoryDropdown from './category-dropdown';
 
 class DropdownGroup extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      categories: [], // we have empty array, that pass to CategoryDropdown
+    };
+    // eslint-disable-next-line no-console
+    console.log(this.state.categories);
+    this.onAddCategory = this.onAddCategory.bind(this);
+  }
+
+  // onAddCategory(newCategory) {
+  //   this.setState({ categories: [...this.state.categories, newCategory] });
+  //   // eslint-disable-next-line no-console
+  //   console.log(this.state.categories);
+  // }
+
+
   onSelectCategory(path) {
     this.props.onChangeEvents(path);
   }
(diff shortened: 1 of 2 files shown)


client/src/components/dropdown-group.js
```diff
@@ -7,6 +7,25 @@ import MyDropdown from './my-dropdown';
...
+  }
+
+  // eslint-disable-next-line no-unused-vars
+  shouldComponentUpdate(nextProps, nextState, nextContent) {
+    if (typeof nextProps.categories !== 'undefined' && typeof this.props.categories !== 'undefined') {
+      if (nextProps.categories.length !== this.props.categories.length) {
+        this.setState({ categories: nextProps.categories });
+      }
+    }
+    return this.state.value !== nextState.value;
```
vitaliimak, line 26: It isn't a good idea to use `if` inside another `if`

vitaliimak (commented):
Please, don't push `console.log`-s in github. It interferes to see code

Don't forget to use current dev branch. If you don't merge dev into your feature branch, your code can work in feature branch, but it will break after merge and you will need to fix new issue. Or maybe this issue is fixed in current dev branch

Did you try to use another React module instead of `rc-menu`?
You get all events when user open `Events page`, did you try to make filtering only in frontend?

narekacy:
> Please, don't push `console.log`-s in github. It interferes to see code

This is just for testing now.

> Don't forget to use current dev branch. If you don't merge dev into your feature branch, your code can work in feature branch, but it will break after merge and you will need to fix new issue. Or maybe this issue is fixed in current dev branch

I remember this. It will work, don't worry. I will merge later, may be another merges will be before mine.

> Did you try to use another React module instead of `rc-menu`?

`rc-menu` was used by another developer. My goal is to deal with 'Reset' button. On the other hand I agree with you due to React.js components philosophy.

> You get all events when user open `Events page`, did you try to make filtering only in frontend?

No, dude :) This is no either my task.

narekacy:
> > Please, don't push `console.log`-s in github. It interferes to see code
> 
> This is just for testing now.
> 
> > Don't forget to use current dev branch. If you don't merge dev into your feature branch, your code can work in feature branch, but it will break after merge and you will need to fix new issue. Or maybe this issue is fixed in current dev branch
> > I remember this. It will work, don't worry. I will merge later, may be another merges will be before mine.
> 
> > Did you try to use another React module instead of `rc-menu`?
> > `rc-menu` was used by another developer. My goal is to deal with 'Reset' button. On the other hand I agree with you due to React.js components philosophy.
> 
> > You get all events when user open `Events page`, did you try to make filtering only in frontend?
> > No, dude :) This is no either my task.

client/src/components/dropdown-group.js
narekacy: This is just for testing. If this works, of course I will make code better then. Thanks

PR #35, opened 2020-03-30
Feature 79 restaurant list voznyarskyy

VozNick:


server/src/main/java/com/lunchtime/seed/DatabaseSeed.java
```diff
@@ -47,94 +58,70 @@ public DatabaseSeed(RestaurantRepository restaurantRepository,
...
-            seedRestaurant();
-        }
-
-        if (feedbackRepository.count() == 0L) {
-            seedFeedback();
-        }
-
+        if (personRepository.count() == 0L) seedPerson();
+        if (restaurantRepository.count() == 0L) seedRestaurant();
+        if (feedbackRepository.count() == 0L) seedFeedback();
```
vitaliimak: image
It's a bad practice - if statement without curly braces

client/src/components/restaurant/restaurant-card-view.js
```diff
@@ -0,0 +1,28 @@
...
+class RestaurantCardView extends Component {
+  render() {
+    const { restaurant } = this.props;
+    const link = `/restaurants/${restaurant.id}`;
+    return (
+      <Card className="m-2" border="dark">
+        <Link to={link}>
+          <Card.Body>
+            <Card.Title className="">{restaurant.name}</Card.Title>
+            <Card.Subtitle className="restaurant-name mb-2">4,6 stars, Hours: 12-24</Card.Subtitle>
```
vitaliimak: Restaurant model has `workingTime` property

server/src/main/java/com/lunchtime/seed/DatabaseSeed.java
```diff
@@ -47,94 +58,70 @@ public DatabaseSeed(RestaurantRepository restaurantRepository,
...
                 restaurantName[i.intValue()].concat("@gmail.com").toLowerCase(),
                 "Rivne, street ".concat("1" + i.toString()),
                 "www.".concat(restaurantName[i.intValue()]).concat(".ua").toLowerCase(),
-                "Description for restaurant ".concat(restaurantName[i.intValue()]),
-                "12-24", false,
-                i + 1L, person.get(i.intValue()), i.intValue() + 1, cordLongitude[i.intValue()],
-                cordLatitude[i.intValue()], Instant.now(),
-                i + 1L, Instant.now(), i + 1L);
-
+                "A short description about restaurant ".concat(restaurantName[i.intValue()]).concat(", and some other info."),
```
vitaliimak: What happen with frontend if description will be normal? E.g. 200 - 300 characters?

vitaliimak: If you create testing data you should add normal (real) information. Not Lorem Ipsum

vitaliimak (changes requested):
image

I've added new Restaurant, but I don't see it in list
image

server/src/main/java/com/lunchtime/seed/DatabaseSeed.java
```diff
@@ -47,94 +58,70 @@ public DatabaseSeed(RestaurantRepository restaurantRepository,
...
             }
         }
-
     }
 
     public void seedRestaurant() {
-
-        List<Person> person = personRepository.findAll();
-
         for (Long i = Long.valueOf(0); i < restaurantName.length; i++) {
```
r-yevchuk, line 67: `Long i` I think this variable might be a primitive int.

r-yevchuk (changes requested):

server/src/main/java/com/lunchtime/seed/DatabaseSeed.java
VozNick: I removed description from cards. I think it's useless in list. I searched through internet and usually there is the most important information to book a restaurant: hours, raiting, address, photo. Because it goes without saying that there are only restaurants. If a user wants to see info, he will open a rest page and will look there full info about it. 

P.S. Nonetheless, my opinion is that not all restaurants will have a big description, maybe some will have few words for example. 

P.S. I think if the main goal is to find were to eat, user usually looks on photos (menu) and working hours, address

client/src/pages/restaurant-list.js
```diff
@@ -1,22 +1,101 @@
...
+    };
+    this.handlePageChange = this.handlePageChange.bind(this);
+  }
+
+  componentDidMount() {
+    this.getAll(this.state.number);
+  }
+
+  getAll(props) {
+    Api.getAll(`restaurants?page=${props}&size=21`)
```
vitaliimak: Try to avoid to use "magic" constants. You can create class property e.g. `pageSize = 21` and use this named constant.

client/src/pages/restaurant-list.js
```diff
@@ -1,22 +1,101 @@
...
+      </Container>
+    );
+  }
+
+  initPagination() {
+    const { number, totalPages } = this.state;
+    if (totalPages === 1) {
+      return;
+    }
+    // eslint-disable-next-line consistent-return
```
vitaliimak: You can return `null` in line 70 to prevent this eslint error

server/src/main/java/com/lunchtime/controllers/RestaurantController.java
```diff
@@ -5,12 +5,12 @@
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
 import com.lunchtime.service.RestaurantService;
 import org.springframework.data.domain.Page;
+import org.springframework.data.domain.PageRequest;
```
vitaliimak: Remove unused import

vitaliimak (commented):
Nice work! Could you also fix 3 small issues?

vitaliimak (dismissed):

r-yevchuk (changes requested):
Background should be full screen
Знімок екрану_2020-04-05_23-31-56

r-yevchuk:
And i think you should rename ListRestaurant link to Restaurants

client/src/pages/restaurant-list.js
```diff
@@ -5,14 +5,15 @@ import ButtonToolbar from 'react-bootstrap/ButtonToolbar';
...
+import '../styles/restaurant-list.css';
 
 class RestaurantList extends Component {
   constructor(props) {
     super(props);
     this.state = {
       totalPages: 0,
       number: 0,
-      pageSize: 21,
+      pageSize: 2,
```
vitaliimak: Is it right number?

r-yevchuk (approved):

vitaliimak (approved):

Melnyk9506 (approved):

PR #36, opened 2020-04-03
Feature 14 add possibility to make order makarchenkov

vitaliimak:
Work:
1. can create new order
2. can add new table (tables have unique numbers in restaurants)
3. can filter available tables in restaurant in specific time period
4. simple form for creating new order

TODO:
1. reload new-order page
2. add default order status
3. startTime mustn't be less then current time
4. finishTime must be more then startTime
5. isDeleted in every requests
6. fix (fetch = FetchType.EAGER)
7. a lot of bugs...

Table
```
{
	"number": 1,
	"capacity": 4,
	"available": true,
	"restaurant": {
		"id": 7
	}
}
```
Order
```
{
	"person": {
		"id": 1
	},
	"startTime": "2020-04-05T16:00:00.720Z",
	"finishTime": "2020-04-05T18:00:00.720Z",
	"visitors": 2,
	"table": {
		"id": 1
	}
}
```
Order status
```
{
	"name": "open"
}
```
Available tables request
`localhost:8080/api/tables/available?id=7&start=2020-4-05 16:00&finish=2020-04-05 19:00`
Feel free to criticize


server/src/main/java/com/lunchtime/service/implementation/OrderServiceImplementation.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public OrderServiceImplementation(
+        OrderRepository orderRepository,
+        RestaurantTableService restaurantTableService,
+        PersonService personService
+    ) {
+        this.orderRepository = orderRepository;
+        this.restaurantTableService = restaurantTableService;
+        this.personService = personService;
+    }
```
narekacy: **[@RequiredArgsConstructor](https://projectlombok.org/features/constructor)** will make code easier (it creates constructor with reqired fields `final` fields in this case).

server/src/main/java/com/lunchtime/service/implementation/OrderServiceImplementation.java
```diff
@@ -0,0 +1,66 @@
...
+
+        Optional<RestaurantTable> table = restaurantTableService.findById(order.getTable().getId());
+        if (table.isPresent()) {
+            order.setTable(table.get());
+        } else {
+            return null;
+        }
+        return orderRepository.save(order);
+    }
+
```
narekacy: Sorry, but I don't undersand lines 34-46 here. Could you explain this, please?

narekacy: I mean, why do you use setters if object is present?

vitaliimak: `Optional` is a class which helps to handle `NullPointerException`. `findById(id)` can don't find object in this case statement `else` will be used

I need to check if objects (`Person and RestaurantTable`) aren't `null` before set them

narekacy: It is Ok that you return null when person or restaurant are null. Otherwise, it seems that you find person from incoming order in line 34, and if it is present, you set it back to this order, isn't it?

vitaliimak: Yes

narekacy: So, do we need to set it back when it exists?

vitaliimak: image
image

Yes

narekacy: Now I got it. Just passing id... 
Thank you!

server/src/main/java/com/lunchtime/service/implementation/OrderServiceImplementation.java
vitaliimak: Thanks! Very  useful!

vitaliimak:
@narekacy @Andrii35 @Melnyk9506 PR is ready to review

client/src/app.js
```diff
@@ -44,7 +45,13 @@ class App extends Component {
...
           <Route path="/about" component={About} />
           <Route path="/events" component={Events} />
-          <Route path="/restaurants/:id" component={Restaurant} />
+          <Route path="/restaurants/:id/new-order" component={NewOrder} />
+          <Route
+            path="/restaurants/:id"
+            render={(routeProps) => {
+              return <Restaurant isAuthenticated={isAuthenticated} {...routeProps} />;
+            }}
+          />
```
narekacy, line 61: I guess we need change something here. When I work with app 'routes' via app buttons, it is ok. But this is what happens with URL when I type it manually:
`...restaurants` - renders all Reastaurants
`...restaurants/2` - renders Restaurant with id 2
`...restaurants/2/new-order` - i get this TypeError:
image

May `BrowserRouter` can help.

vitaliimak: Fixed. It is SPA, use normal routing

server/src/main/java/com/lunchtime/repository/OrderRepository.java
```diff
@@ -0,0 +1,41 @@
...
+public interface OrderRepository extends JpaRepository<Order, Long> {
+
+    @Query("select o from Order o where (o.isDeleted = false) and "
+        + "(o.table.restaurant.id = :id) and "
+        + "((:start >= o.startTime and :start < o.finishTime) or "
+        + "(:finish > o.startTime and :finish < o.finishTime) or "
+        + "(:start <= o.startTime and :finish >= o.finishTime))")
+    List<Order> findAllOrdersByRestaurantIdInTime(
+        @Param("id") Long id, @Param("start") Date start, @Param("finish") Date finish);
+
```
narekacy, line 25: Can't understant line's 22 purpose after 20-12 lines. 

And why we can't use this in line 21?  `...:finish <= o.finishTime`

server/src/main/java/com/lunchtime/seed/DatabaseSeed.java
```diff
@@ -55,26 +60,8 @@
...
-        this.restaurantRepository = restaurantRepository;
-        this.feedbackRepository = feedbackRepository;
-        this.personRepository = personRepository;
-        this.dishRepository = dishRepository;
-        this.categoryFoodRepository = categoryFoodRepository;
-        this.menuItemDishRepository = menuItemDishRepository;
-        this.eventCategoryRepository = eventCategoryRepository;
-        this.eventRepository = eventRepository;
-
-    }
```
narekacy, line 77: Good job!

server/src/main/java/com/lunchtime/repository/OrderRepository.java
vitaliimak: Line 22 new Order - from 13:00 to 18:00. If there is exist active order from 14:00 to 17:00. Than new Order can't be created

Line 21 new Order - from 13:00 to 18:00. Exist order - from 17:00 to 18:05

narekacy: > Line 21 new Order - from 13:00 to 18:00. Exist order - from 17:00 to 18:05
may be you mean this in line 20?

> Line 22 new Order - from 13:00 to 18:00. If there is exist active order from 14:00 to 17:00. Than new Order can't be created
So we need `!not` operator? Or what? :)

vitaliimak: > Line 21 new Order - from 13:00 to 18:00. Exist order - from 17:00 to 18:05
> may be you mean this in line 20?

start = 13:00 < (o.startTime == 17:00)
Line 21

> So we need !not operator? Or what? :)

It works well now, but you can change query and try

server/src/main/java/com/lunchtime/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,83 @@
...
+        if (table.isPresent() && table.get().getCapacity() >= order.getVisitors()) {
+            order.setTable(table.get());
+        } else {
+            return null;
+        }
+
+        Optional<OrderStatus> status = orderStatusService.getOrderStatusByName(newOrderStatus);
+        status.ifPresent(order::setStatus);
+
+        return orderRepository.save(order);
```
narekacy, line 64: What app should do if status isn't present?

server/src/main/java/com/lunchtime/service/impl/OrderServiceImpl.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public Order deleteOrder(Long id) {
+        return getOrderById(id)
+            .map(order -> {
+                order.setDeleted(true);
+                return orderRepository.save(order);
+            })
+            .orElseGet(null);
+    }
```
narekacy, line 82: `.orElse(null)` here (and another places like this) will be enough.

- orElse(): returns the value if present, otherwise return other
- orElseGet(): returns the value if present, otherwise invoke other and return the result of its invocation

Good and short explanation **[here](https://www.baeldung.com/java-optional-or-else-vs-or-else-get#why%20bother)**

server/src/main/java/com/lunchtime/service/impl/OrderServiceImpl.java
vitaliimak: Nothing

narekacy: Due to impossibility of status absense, I find we can use simple `OrderStatus` here.

server/src/test/java/com/lunchtime/service/impl/OrderServiceImplTest.java
```diff
@@ -0,0 +1,217 @@
...
+
+        when(orderRepository.findAllOrdersByTableInTime(order.getTable().getId(), startDate, finishDate))
+            .thenReturn(Collections.emptyList());
+        when(orderRepository.save(order)).thenReturn(order);
+
+        Order createdOrder = orderServiceImpl.saveOrder(order);
+
+        assertNull(createdOrder);
+    }
+}
```
narekacy, line 200: I'm not close to testing. What I see logically, test method name should be easier and clear for its reader. Use comments to exaplain what test's purpose is.

Then, it seems that Date can be created easily and moved to `setup()`. If we'll need another date in our test, we can just modify it whereas, as well as `order`. 

`when()` statements are repeated multiply. I guess there is way to improve it. 

I will discover Testing more soon and give you further feedback. Also I'm interested in team's developers opinions about this testing.

vitaliimak: > Use comments to exaplain what test's purpose is.

Absolutely no! Don't use comments in code, your code must be selfdocumented. And in test cases, method names are documentation of project

server/src/main/java/com/lunchtime/service/impl/OrderServiceImpl.java
vitaliimak: I don't understand

narekacy: Do we can use smth like this?

```
OrderStatus status = orderStatusService.getOrderStatusByName(newOrderStatus);
order.setStatus(status);
```

server/src/main/java/com/lunchtime/models/Order.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @NotNull
+    @Temporal(TemporalType.TIMESTAMP)
+    @Column(name = "start_time")
+    private Date startTime;
+
+    @NotNull
+    @Temporal(TemporalType.TIMESTAMP)
+    @Column(name = "finish_time")
+    private Date finishTime;
```
narekacy, line 44: As I discovered `@Temporal` annotation by default sets `TemporalType.TIMESTAMP`. Anyway, I can't understand clearly its benefit. Could you explain simply the purpose of `@Temporal` annotation here and how this is used in rest part of code now? As I understand, this is for converting incoming date from UI to sql type in DB?

server/src/main/java/com/lunchtime/models/Order.java
```diff
@@ -0,0 +1,83 @@
...
+    @Column(name = "visitors")
+    private Integer visitors;
+
+    @NotNull
+    @ManyToOne(fetch = FetchType.EAGER)
+    @JoinColumn(name = "table_id", referencedColumnName = "id")
+    private RestaurantTable table;
+
+    @Column(name = "dish_details_id")
+    private Long dishDetails;
```
narekacy, line 64: Can't find out what should be `dishDetails`. It should be dish/dishes that user ordered (or not ordered) while booked table, shouldn't it?

server/src/main/java/com/lunchtime/service/impl/OrderServiceImpl.java
narekacy: May be we need `@ColumnDefault` annotation in `OrderStatus` entity and nothing initialized here?
```
@NotNull
@Size(max = 32)
@Column(name = "name")
@ColumnDefault("new")
private String name;
```

vitaliimak: `getOrderStatusByName` returns `Optional`
It is hardcode logic, in next tasks we will add possibility to change order statuses
By default, seed file generates all statuses what we need

client/src/components/order/new-order.js
```diff
@@ -0,0 +1,279 @@
...
+    });
+  }
+
+  formatDateTime(date) {
+    return `${date.getFullYear()}-${date.getMonth() + 1}-${date.getDate()} ${date.getHours()}:${date.getMinutes()}`;
+  }
+
+  isDateTimesInvalid() {
+    return this.state.startDate >= this.state.finishDate;
+  }
```
narekacy, line 144: I think it is better to do this check when 'Create new order' button is clicked. I tried this and find UX can be better. 

Consider you pick start date and time for next week's Saturday on 8 p.m. and available finish date will be also next Saturday and starting 8:15 pm. This will let to alert '_Finish time should be **after** start time!_' only when user wants to pick finish date that is 'before' start date.

server/src/main/java/com/lunchtime/models/Order.java
vitaliimak: Yes, may be)

server/src/main/java/com/lunchtime/service/impl/OrderServiceImpl.java
narekacy: As i know `@ColumnDefault("new")` is default value for any new order. I'm not sure if it is really hardcoded for this purpose. Also it can be changed any time

server/src/main/java/com/lunchtime/models/Order.java
narekacy: So I guess this may be temporarily decision to persist them in `Long` type?

server/src/main/java/com/lunchtime/models/Order.java
vitaliimak: I think, it converts Java Date to Database Date. I can't explain more :(

narekacy: Ok, thanks

client/src/components/order/new-order.js
vitaliimak: Button `Save order` is disable when form is invalid. User can't click it when form is invalid

narekacy: I was mistaked. I had to write this:

> I think it is better to do this check when **inactive finish date on datepicker** is choosen before start date. I tried this and find UX can be better.

server/src/test/java/com/lunchtime/service/impl/OrderServiceImplTest.java
narekacy: `personCanMakeOrderInSpecificTimeIfRestaurantTableIsAvailable` <- this is not the best way to explain what test method does. Also it is recommended to use only one `Mock` in testing class

client/src/components/order/new-order.js
vitaliimak: > finish date on datepicker is choosen before start date

I don't understand. And in any cases, `finish date` must be chose only after `start date`. People think from start to finish not vice versa.

server/src/main/java/com/lunchtime/service/impl/OrderServiceImpl.java
vitaliimak: Do you propose to remove OrderStatus entity, service, controller and seed function?

server/src/test/java/com/lunchtime/service/impl/OrderServiceImplTest.java
vitaliimak: Why? What do you don't understand?

> Also it is recommended to use only one Mock in testing class

I'm newby in testing. Could you show examples of right solutions?

server/src/main/java/com/lunchtime/models/Order.java
vitaliimak: Yes

server/src/test/java/com/lunchtime/service/impl/OrderServiceImplTest.java
vitaliimak: My business logic depends on all those Mocks, how can I should use all necessary repositories without Mock?

vitaliimak: > Then, it seems that Date can be created easily and moved to setup(). If we'll need another date in our test, we can just modify it whereas, as well as order.

I've tried to do it, but it doesn't work

server/src/main/java/com/lunchtime/service/impl/OrderServiceImpl.java
vitaliimak: Done

vitaliimak:
@Andrii35 @Melnyk9506 I've fixed merge conflicts, could you review PR?

client/src/components/order/new-order.js
narekacy: I'm not what are you talking about. I'm not saying that _People think from start to finish not vice versa._ I mean alerts, we should change their behavior. 

> Consider you pick start date and time for next week's Saturday on 8 p.m. 

**now alert "Finish time should be more than Start time" is shown**. But: user haven't choosen finish date yet. 

Imagine that we have no any alert in case above.

**Case 1:** Then, user chooses available finish date - also next Saturday and starting 8:15 pm. It is ok due to time rules and user won't see any alerts anymore.

**Case 2:** But, when he wants to choose finish date that in fact is before his choosen start date, here we should show **alert**. Moreover, for better UX we should make dates 'before' choosen start inactive, but still show alert if user tries to pick them.

Thus, it will let to alert "Finish time should be after start time!" only in one case - when user wants to pick finish date that is 'before' start date.

server/src/main/java/com/lunchtime/service/impl/OrderServiceImpl.java
narekacy: I am surprised for your question. Why do you think I do that proposal for you, if I haven't?

server/src/test/java/com/lunchtime/service/impl/OrderServiceImplTest.java
narekacy: I'm also new in testing, so I can be mistaken. **[This](https://stackoverflow.com/a/17810004/13334444)** tells us that:
_According to the principle of Test only one thing per test, there may be several stubs in one test, but generally there is only one mock._

> I've tried to do it, but it doesn't work

**[Try](https://examples.javacodegeeks.com/core-java/junit/junit-setup-teardown-example/)** `tearDown()` to do a prepration and post clean up process for each of the JUnit method test call.

server/src/main/java/com/lunchtime/models/Order.java
narekacy: I think it is better to write this near `Long` usage so that other developers could understand it.

narekacy:
Your code is working, it is ok and I will approve it now for our demo tomorrow. 

However, you use multiple times `!important` in this PR and **[another](https://github.com/Social-projects-Rivne/Lunch_Time/pull/61/commits/678ea7896016208e109042e72a183f9bd9b8e26c)** one. This can be ok for our demo presentation but should be changed as `!important` usage is really **[bad](https://stackoverflow.com/a/3706876/13334444)** practice.

narekacy (dismissed):

server/src/main/java/com/lunchtime/service/impl/OrderServiceImpl.java
vitaliimak: If you want to leave only "new" status, than other functionality aren't necessary, isn't?
`@ColumnDefault("new")` - is a bad idea, because of we can use "open" instead of "new", or other word

server/src/main/java/com/lunchtime/models/Order.java
vitaliimak: What?

client/src/components/order/new-order.js
vitaliimak: Current alert teach user how to use our application.

> Then, user chooses available finish date

What happen if user doesn't choose finish date? He or she doesn't see any alert and doesn't click Save button (form is invalid)

> Moreover, for better UX we should make dates 'before' choosen start inactive

I couldn't do it in this datapicker https://github.com/Hacker0x01/react-datepicker/issues/2077 and other

server/src/main/java/com/lunchtime/service/impl/OrderServiceImpl.java
narekacy: 1. Do we can change `@ColumnDefault("new")` value after it is persisted in DB?
2. As I understand, all orders created can have only one status until this order can be managed by restaurant (or user) ?

server/src/main/java/com/lunchtime/models/Order.java
narekacy: Something like comments, or documentation.

client/src/components/order/new-order.js
narekacy: It is clear that alert should be shown if smth goes wrong. When user picks start date, he hasn't done smth wrong to see alert. We can move finish date due to choosen start date and + 15 minutes, for example. Now user can understand that he isn't able to use finish time 'before' start. 

Also dates before finish time should be inactive and we shall do it even in another, special task. Otherwise we allow user to do abnormal thing in our app.

Also we shold restrict user to make order starting 20.04.20 and ending 01.05.20 as restaurant will be closed from 24:00 to 12:00, for example.

This should be marked to fix in another tasks or may be this.

server/src/main/java/com/lunchtime/models/Order.java
vitaliimak: I don't write useless comments. If developer want to know what is it, he or she can open database scheme

image

narekacy: So, you don't use useless comments but create useless actions for another developer? You told this is temporary desicison to use `Long` here.

client/src/components/order/new-order.js
vitaliimak: > We can move finish date due to choosen start date and + 15 minutes, for example. 

I've tried to do it, but couldn't

> Also dates before finish time should be inactive

...
> I couldn't do it in this datapicker Hacker0x01/react-datepicker#2077 and other

> Also we shold restrict user to make order starting 20.04.20 and ending 01.05.20 as restaurant will be closed from 24:00 to 12:00, for example.

It's a good idea, I think, you should create new task

narekacy: > Also dates before finish time should be inactive

Sorry, it should be this:
Also dates and times before start time should be inactive.

> It's a good idea, I think, you should create new task

We can create this task only whe this branch will be merged.

vitaliimak: > Sorry, it should be this:
Also dates and times before start time should be inactive.

...
> I couldn't do it in this datapicker Hacker0x01/react-datepicker#2077 and other

narekacy: According to this inactive 'days' presense, I guess it is possible to implement inactive status. Let us do it in another task.

image

vitaliimak: Did you read issue? Making date is inactive is possible, but not time

server/src/test/java/com/lunchtime/service/impl/OrderServiceImplTest.java
vitaliimak: Thanks! Very useful annotation `@After`, I couldn't find it earlier
But, I can't change all Mocks to Stubs

client/src/components/order/new-order.js
narekacy: In your code not only time but date can be choosen earlier than start date.
We can fix time but still can choose finish date before start, so we need to fix both.

vitaliimak: Server handles this case and client shows message if request is wrong. It isn't problem

server/src/test/java/com/lunchtime/service/impl/OrderServiceImplTest.java
narekacy: I'm also learning testing at the moment. Minute ago I found **[this](https://stackoverflow.com/questions/3459287/whats-the-difference-between-a-mock-stub)**, it may help us.

Andrii35 (approved):
I will approve it. 
Functionality works.

Melnyk9506 (approved):

narekacy (approved):

PR #37, opened 2020-04-05
Feature 22. Leave feedback to restaurant as a user. narek

narekacy:
Need to fix:

**in [send.js](https://github.com/Social-projects-Rivne/Lunch_Time/blob/41622653b184bf569da070ebd84783e14673c893/client/src/components/feedback/send.js#L114)**
- get response status code in **`sendApi()`** method


client/src/components/feedback/send.js
```diff
@@ -0,0 +1,220 @@
...
+    this.cancel = this.cancel.bind(this);
+    this.showFeedbackSent = this.showFeedbackSent.bind(this);
+    this.showFeedbackNotSent = this.showFeedbackNotSent.bind(this);
+    this.showTimer = this.showTimer.bind(this);
+  }
+
+  onSubmit() {
+    if (this.state.description.length < 10) {
+      this.tooShortFeedback();
+    } else if (this.state.description.length > 400) {
```
vitaliimak: Input has attributes min- and maxlength https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefmaxlength

client/src/components/feedback/send.js
```diff
@@ -0,0 +1,220 @@
...
+      inputDisabled: true,
+      currentCallId,
+    });
+    setTimeout(() => {
+      if (currentCallId !== this.state.currentCallId) return;
+      this.setState({
+        showTimer: false,
+      });
+    }, 4000);
+    setTimeout(() => {
```
vitaliimak, line 136: I think it is bad practice to use `setTimeout`. It makes code more difficult and unpredictable

client/src/components/feedback/send.js
```diff
@@ -0,0 +1,220 @@
...
+        restId: this.props.id,
+        description: this.state.description,
+      })
+        .then((response) => {
+          if (currentCallId !== this.state.currentCallId
+            && !this.state.isLoading
+            && response) return;
+          this.showFeedbackSent();
+        })
+        .catch((error) => {
```
vitaliimak, line 154: Catch must be at the end

client/src/components/feedback/send.js
```diff
@@ -0,0 +1,220 @@
...
+    });
+    setTimeout(() => {
+      if (currentCallId !== this.state.currentCallId) return;
+      this.setState({
+        showTimer: false,
+      });
+    }, 4000);
+    setTimeout(() => {
+      if (currentCallId !== this.state.currentCallId) return;
+      Api.post('feedback', {
```
vitaliimak, line 138: image

server/src/main/java/com/lunchtime/controllers/FeedbackController.java
```diff
@@ -1,24 +1,39 @@
...
-    private final FeedbackServiceImplement feedbackServiceimplementation;
+    private final FeedbackServiceImpl feedbackServiceimplementation;
 
-    public FeedbackController(FeedbackServiceImplement feedbackServiceimplementation) {
+    public FeedbackController(FeedbackServiceImpl feedbackServiceimplementation) {
         this.feedbackServiceimplementation = feedbackServiceimplementation;
     }
 
+    @PostMapping
+    public ResponseEntity<FeedbackDto> create(@RequestBody FeedbackDto feedbackDto) throws URISyntaxException {
```
vitaliimak: You should handle wrong request data (don't believe frontend users) Server should return BedRequest if user sends wrong data. Server shouldn't throw error 500

client/src/components/feedback/feedback.js
```diff
@@ -51,19 +50,16 @@ class Feedback extends Component {
 
   render() {
```
vitaliimak, line 97: image
I've created new feedback, but I don't see it

client/src/components/feedback/send.js
```diff
@@ -0,0 +1,220 @@
...
+    this.setState({
+      isLoading: false,
+      showTimer: false,
+      inputDisabled: false,
+    });
+  }
+
+  showFeedbackSent() {
+    this.setState({ description: '', feedbackSent: true });
+    setTimeout(() => {
```
vitaliimak, line 176: Too many setTimeotes

client/src/components/shared/my-alert.js
```diff
@@ -0,0 +1,26 @@
+import React, { Component } from 'react';
+import { Alert } from 'react-bootstrap';
+import PropTypes from 'prop-types';
+
+class MyAlert extends Component {
```
vitaliimak: Nice! Creating shared component when it really needs!

client/src/components/shared/timer.js
```diff
@@ -0,0 +1,50 @@
...
+
+  hide() {
+    this.props.showTimer(false);
+  }
+
+  doIntervalChange() {
+    if (this.state.count === 0) {
+      this.hide();
+    }
+    this.myInterval = setInterval(() => {
```
vitaliimak, line 28: `setInterval()` It looks like you try to write the most difficult code in the world. Why don't you just do your simple task?

client/src/services/api.js
```diff
@@ -34,10 +34,19 @@ class Api {
     } catch (error) {
       return null;
     }
-
     return data;
   }
 
+  post(endpoint, body) {
+    return axios.post(`${this.apiUrl}${endpoint}`, body)
```
vitaliimak, line 41: What happen if other developer sends `endpoint = '/person'`?
You should use url handler function, which handles last slash

server/src/main/java/com/lunchtime/models/Feedback.java
```diff
@@ -25,7 +25,7 @@
 
     @NotNull
     @Column(name = "date")
-    private Date date;
+    private Instant instant;
```
vitaliimak: image

What mean this name? How should other developers, especially API users (frontend developers),  understand this name? 
And also, don't call variables as their type. `Date date, Number number, Boolean boolean` Those names don't explain what is variable mean

server/src/main/java/com/lunchtime/service/dto/FeedbackDto.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @NotNull
+    private long restId;
+
+    private boolean isActive;
+
+    private Instant instant;
+
+    @NotNull
+    private long personId;
```
vitaliimak: We've decided to use Long as id's

server/src/main/java/com/lunchtime/service/impl/EventServiceImpl.java
```diff
@@ -1,4 +1,4 @@
-package com.lunchtime.implementation; //TODO should be in service package
+package com.lunchtime.service.impl; //TODO should be in service package
```
vitaliimak: As you fix this TODO, you can remove it

server/src/main/java/com/lunchtime/service/impl/FeedbackServiceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+            feedbackDto.setId(feedback.getId());
+            feedbackDto.setActive(feedback.getIsActive());
+            feedbackDto.setCounterDislike(feedback.getCounterDislike());
+            feedbackDto.setCounterLike(feedback.getCounterLike());
+            feedbackDto.setInstant(feedback.getInstant());
+            return feedbackDto;
+        } catch (EntityNotFoundException e) {
+            e.printStackTrace();
+        }
+        return null;
```
vitaliimak: It looks like DTO makes code more difficult

server/src/test/java/com/lunchtime/controllers/FeedbackControllerTest.java
```diff
@@ -0,0 +1,84 @@
...
+    private FeedbackController feedbackControllerUnderTest;
+
+    @Before
+    public void setUp() {
+        initMocks(this);
+        feedbackControllerUnderTest = new FeedbackController(mockFeedbackServiceImpl);
+    }
+
+    @Test
+    public void testCreate() throws Exception {
```
vitaliimak, line 39: Test names should be self documented. Your colleagues should read test name and understand what tested method or class does

server/src/test/java/com/lunchtime/controllers/FeedbackControllerTest.java
```diff
@@ -0,0 +1,84 @@
...
+        feedbackDto.setCounterDislike(0);
+
+        final ResponseEntity<FeedbackDto> result = feedbackControllerUnderTest.create(feedbackDto);
+
+        if(result != null) {
+            verify(mockFeedbackServiceImpl, times(1)).save(feedbackDto);
+            verifyNoMoreInteractions(mockFeedbackServiceImpl);
+        } else {
+            throw new Exception();
+        }
```
vitaliimak: You should use `assert` statements
https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html

vitaliimak (commented):
Could you rename branch according to us standard?

vitaliimak:
> default values are not genereted in lines 22, 40, 44

image
It works

client/src/components/feedback/send.js
narekacy: Thank you!

client/src/components/feedback/send.js
narekacy: Hmm.. Why difficult and unpredictable? To be honest it was easy for me. 
If you have an examples, I will rewrite my code with pleasure. And what alternatives to `setTimeout` are to use in this functionality?

server/src/main/java/com/lunchtime/service/dto/FeedbackDto.java
narekacy: Don't forget that `personId` is temorary solution here while we have no authorization completed. **[Also](https://forums.asp.net/post/5572845.aspx)** it is advised to use primitives in DTO classes.

narekacy:
> Could you rename branch according to us standard?

F -> Feature?

client/src/components/shared/timer.js
narekacy: My experience does't let me to do such difficult code yet :) This is a sample for Timer in React.js. But, if there any easier ways to do this, just give me a hint. Thus, I wil be able to refactor this code.

client/src/components/feedback/send.js
narekacy: Actually, this is for response handling in `sendApi()` method in `send.js`. Id is valid, and you can check it in another request, that gives 201 status with the same id. Just give me a few minutes, I will upload new commits.

narekacy: I have written about this in draft PR description (response handling).

client/src/components/feedback/send.js
narekacy: I am not sure. Lines 125-129 must be completed. That's why i situated them after 'possible' catch case.

client/src/components/feedback/feedback.js
narekacy: Yes, this is another issue. Unfornutately, we can't contact the developer who wrote this code. I tried a lot, and now I can't figure out how to deal with it. This also was written in this PR description.

server/src/main/java/com/lunchtime/service/impl/EventServiceImpl.java
narekacy: Just forgot this :) Thanks

client/src/components/feedback/send.js
narekacy: Can't find another way yet

server/src/main/java/com/lunchtime/controllers/FeedbackController.java
narekacy: Thanks!

client/src/services/api.js
narekacy: Thanks!

server/src/test/java/com/lunchtime/controllers/FeedbackControllerTest.java
narekacy: Thanks!

server/src/test/java/com/lunchtime/controllers/FeedbackControllerTest.java
narekacy: Thanks!

server/src/main/java/com/lunchtime/models/Feedback.java
narekacy: Thanks!

narekacy:
> > default values are not genereted in lines 22, 40, 44
> 
> image
> It works

I've set them manually in `FeedbackServiceImpl` now. Just give me 2 minutes and I will update this PR, you can see the same in `FeedbackMapper` class (I mean manually set).

client/src/components/feedback/send.js
vitaliimak: If you need to invoke some necessary code (close DB connection, close file, etc) you can use `.finally()`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

vitaliimak: https://habr.com/en/company/mailru/blog/269465/

client/src/components/feedback/send.js
vitaliimak: Async function inside async function inside async function... Too many setTimeouts, too hard. But may be other team's members think different

client/src/components/feedback/send.js
narekacy: 🤣 forgot about finally)
BTW, in some cases it is better to use try-with-recources instead of finally. You mentioned to close DB connection or fille - I mean this.

client/src/components/feedback/send.js
narekacy: Let us consider this, it may be interesting. Even though, I have no other alternatives yet.

client/src/components/feedback/send.js
vitaliimak: There is another way. Your task doesn't force you to create this complicated logic

narekacy: But it still works fine

PR #38, opened 2020-04-06
Fix 98 events seed yevchuk

r-yevchuk:


server/src/main/java/com/lunchtime/models/Event.java
```diff
@@ -54,9 +54,8 @@
     @Column(name = "description")
     private String description;
 
-    @NotNull
     @Column(name = "is_active")
-    private Boolean isActive; //TODO primitive type
+    private boolean isActive;
```
vitaliimak: This name misleads. You use this field to set Event as soft deleted, but name doesn't explain it
I thought that this field helps to filter new Events (which will happen in the future)

r-yevchuk: Fixed

vitaliimak (approved):

VozNick (approved):

Andrii35 (approved):
Ok!

Melnyk9506 (approved):

PR #40, opened 2020-04-08
Feature 22. Leave feedback to restaurant as user. narek

narekacy:
Sorry for difficult code in `send.js` and `feedback.js`. I promise to study **React.js** better and refactor this code later :)


client/src/components/feedback/send.js
```diff
@@ -0,0 +1,260 @@
...
+    this.setState(() => {
+      this.setState({
+        validInput: false,
+      });
+    });
+  }
+
+  validateDescription() {
+    const { description } = this.state;
+    const regex = /(^[a-zA-Z1-9]+[a-zA-Z1-9. ]+)$/;
```
Andrii35: When I insert a coma into a message, a warning message appears and I cannot press the button "Send". In the warning message stay, that i can use coma.

narekacy: It is very strange, as in my last commit 'comma' was used. 

Fixed it again. Thanks

client/src/components/feedback/feedback.js
```diff
@@ -0,0 +1,144 @@
...
+            </span>
+          )}
+        </div>
+      </Container>
+    );
+  }
+}
+
+Feedback.propTypes = {
+  id: PropTypes.string.isRequired,
```
Andrii35, line 141: why there is a string, id must be a number?

narekacy: Try to launch this code with 'number'. It doesn't work. 

We need to pass props as a number. Everything I tried didn't work. Or may be I couldn't do that.

client/src/components/feedback/feedback.js
```diff
@@ -0,0 +1,144 @@
...
+
+  refresh() {
+    this.componentDidMount();
+  }
+
+  render() {
+    const stateFeed = this.state;
+    return (
+      <Container className="feedback">
+        <FeedbackSend id={this.props.id} refreshed={this.refresh} />
```
Andrii35, line 101: Is id here a restaurant id? if so, you would be enough id={this.props} like in the line 26 (const { id } = this.props;)

narekacy: Unfortunately, it doesn't work with only this.props. I tried twice. Could you try on your device?

At all it should work if only we catch id in FeedbackSend later. But it is the same code just in another component.

client/src/components/feedback/send.js
```diff
@@ -0,0 +1,260 @@
...
+    setTimeout(() => {
+      if (currentCallId !== this.state.currentCallId) return;
+      this.setState({
+        showTimer: false,
+      });
+    }, 4000);
+    setTimeout(() => {
+      if (currentCallId !== this.state.currentCallId) return;
+      Api.post('feedback', {
+        personId: 555,
```
Andrii35, line 139: As I understand, Feedback can only send a registered user? I changed the personId to those that I have in the database, then the feedback is sent.

narekacy: Yes. This hardcoded solution is just till we finish our authorization.

client/src/components/feedback/send.js
```diff
@@ -0,0 +1,260 @@
...
+    super(props);
+    this.state = {
+      description: '',
+      isLoading: false,
+      timerCount: 3,
+      showTimer: true,
+      tooShortFeedback: false,
+      tooLongFeedback: false,
+      feedbackSent: false,
+      feedbackNotSent: false,
```
Andrii35, line 20: I think you can make here one variable isSent.

narekacy: Unfortunately, I can't. This is about absense of `null` value of our `bool`. These two `bool` are need to show or not to show two different 'Badges' in order to components state. Initial state of only one bool should be false as we have nothing to show to user yet (while attemt to send feedback is made).

It is possible to connect this with 'attemtCount' bool, but logic will be more complicated.

client/src/components/feedback/send.js
Andrii35: It works now, but why other punctuation marks are not allowed?

client/src/components/feedback/send.js
```diff
@@ -0,0 +1,260 @@
...
+    this.sendApi = this.sendApi.bind(this);
+    this.cancel = this.cancel.bind(this);
+    this.showFeedbackSent = this.showFeedbackSent.bind(this);
+    this.showFeedbackNotSent = this.showFeedbackNotSent.bind(this);
+    this.showTimer = this.showTimer.bind(this);
+  }
+
+  onSubmit() {
+    if (this.state.validInput) {
+      if (this.state.description.length < 10) {
```
vitaliimak, line 38: Why feedback "Cool!" or "+1" aren't valid?

client/src/components/feedback/send.js
narekacy: Our 'zakazchik' doesn't tell us what symbols he wants to be allowed here, so I decided to keep it cheap and strong :)

client/src/components/feedback/send.js
narekacy: I used 'BlaBlaCar' experience, where feedback can be no less than 10 symbols. It is thought that user should write such feedback that can be really like feedback, e.g. "_I've celebrated my birthday here. Everything was prepared due to our time. Very good, I recommend_".

vitaliimak: It isn't a good idea to limit clients in there rights. The better way is to make `description` field is required but let clients write what they want. Some people don't like to write a lot of text
Also our clients can evaluate feedback (set like or dislike)

narekacy: I agree with you. On the other hand, I read somewhere (sadly, can't find it now) that YouTube is indexing and giving attention to feedbacks containing 3 or more words. Yes, we are not YouTube, but feedback should have some context and 'information' for other people. And this can be reached by some kind of description. IMO, feedback should be not less that 10 symbols at least. 

Finaly, we can contact our requirement manager and check what is needed for this feature :)

PR #41, opened 2020-04-08
Fix 91 fix reset button makarchenkov

vitaliimak:


Andrii35 (dismissed):
It works well

narekacy (changes requested):
It works Ok. But, when in refreshes info in 'Categories', we need 'Select' button no more. We can't reach its functionality. So we need to remove this button from UI.

vitaliimak:
@narekacy thanks for review! You are right, I've broken previous functionality. So I've returned possibility to use button `Select` in `Categories`

narekacy (approved):

r-yevchuk (approved):

Melnyk9506 (approved):

Melnyk9506:
works well

PR #46, opened 2020-04-11
Fix issues feature 2 makarchenkov

vitaliimak:


client/src/app.js
```diff
@@ -12,26 +12,51 @@ import NoMatch from './pages/no-match';
...
           <Route path="/events" component={Events} />
           <Route path="/restaurants/:id" component={Restaurant} />
           <Route path="/restaurants" component={RestaurantList} />
-          <Route path="/login" component={Login} />
+          <Route
+            path="/login"
+            render={(routeProps) => {
+              return <Login loginHandler={() => { this.handleLogin(); }} {...routeProps} />;
+            }}
+          />
```
narekacy, line 53: Could you explain how lines 48-53 are working, please?

vitaliimak: https://reacttraining.com/react-router/web/api/Route/render-func

If user go to '/login' arrow function on line 50 return `<Login />` component with default props `routerProps` (with history, match, etc) and also I've added `loginHandler` which return `this.handleLogin()` wher `this === App`

So after user login on page `Login`  function `this.handleLogin()` will be invoked and state will be changed and all components with `isAuthenticated` will be updated

narekacy: Thanks. But as I understand, each time user goes to `/login` page, `this.state.isAuthenticated` will become true each time, won't it?

vitaliimak: If localStorage contains token than `isAuthenticated` will be `true` in other case - `false`

narekacy: I mean line 33 here is invoked (and sets `isAuthenticated ` true) when user wants to go `/login` page

vitaliimak: > after user **login** on page Login function this.handleLogin() will be invoked

https://github.com/Social-projects-Rivne/Lunch_Time/pull/46/files#diff-77bddf7173ca118d8ca26a36e968dc3eR40

narekacy: Does `this.handleLogin()` in line 51 is invoked or passed to `<Login>` component?

vitaliimak: On line 51 I pass arrow function to `Login`. When user successfully login in `Login` component this arrow function will be invoked and invoke `this.handleLogin()` method.
You can add console.log after line 35 https://github.com/Social-projects-Rivne/Lunch_Time/pull/46/files#diff-77bddf7173ca118d8ca26a36e968dc3eR35 and see what object `loginHandler` is
`console.log('Login -> login -> loginHandler', loginHandler);`

PR #49, opened 2020-04-13
Feature 22. Leave feedback to restaurant as user. narek

narekacy:
Merged dev. Check 'send a feedback for restaurant' and previous 'aitenticate' feature. 

Also improved regex.


server/src/main/java/com/lunchtime/controllers/FeedbackController.java
```diff
@@ -1,34 +1,45 @@
...
 import com.lunchtime.models.Feedback;
-import com.lunchtime.service.impl.FeedbackServiceImpl;
+import com.lunchtime.service.FeedbackService;
+import lombok.RequiredArgsConstructor;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.*;
 
-//TODO remove redundant lines
+import javax.validation.Valid;
 @RestController
```
VozNick, line 15: needs to be empty line)
PS. checkstyle error

Andrii35 (dismissed):

server/src/main/java/com/lunchtime/controllers/FeedbackController.java
```diff
@@ -1,34 +1,45 @@
...
 public class FeedbackController {
 
-    //TODO rename variable name to normal CamelCase
-    private final FeedbackServiceImpl feedbackServiceimplementation;
+    private final FeedbackService feedbackService;
 
-    public FeedbackController(FeedbackServiceImpl feedbackServiceimplementation) {
-        this.feedbackServiceimplementation = feedbackServiceimplementation;
+    @PostMapping
+    public ResponseEntity<FeedbackDto> saveFeedback(@Valid @RequestBody FeedbackDto feedbackDto) throws URISyntaxException {
```
VozNick: line to long, checkstyle error

narekacy: I had no warning. Fixed yet.

Andrii35: You will laugh, but the checkstyle indicates that the parenthesis should be immediately after the method name. " /app/src/main/java/com/lunchtime/controllers/FeedbackController.java:24: 'method def lparen' has incorrect indentation level 8, expected level should be 4. [Indentation]" and "/app/src/main/java/com/lunchtime/controllers/FeedbackController.java:24:9: '(' should be on the previous line. [MethodParamPad]"

narekacy: Fixed

VozNick (approved):

Andrii35 (approved):

vitaliimak (commented):
image

There is some strange video artifact. Some flash
https://www.loom.com/share/94aacfcd58d34c4fae63a9c3187d44db

vitaliimak (changes requested):
image
image

narekacy:
@vitalii-mak, actually, I can't implement `shouldComponentMount()` yet, but If you know how it to do, I will learn it with pleasure

As for regex, I think it is better to keep some punctual policy here. If we let users to do what you are asking for, we increase risk that comments can be non-normal syntaxed

VozNick:
Does this PR is going to be merge?

server/src/main/java/com/lunchtime/controllers/FeedbackController.java
Melnyk9506: checkstyle error still have please fix it

Melnyk9506 (approved):
i think this is a working option but we will need to create bug fixes because if you are not logged in then you basically should not see cancel button , you will also need to fix the moment with text input so that you can enter different characters in different languages and especially symbol for switching to another line

vitaliimak (commented):
Hi

vitaliimak (approved):
(

PR #51, opened 2020-04-14
Fix 22. Leave feedback to restaurant as user. narek

narekacy:
Test pull request. Add check if user is logged in to have a possibility to send a feedback to restaurant.


Melnyk9506:
i think this feature will be over when the guest will get some message that he is not authorized or he will have no button to send feedback

narekacy:
> i think this feature will be over when the guest will get some message that he is not authorized or he will have no button to send feedback

Right! He even shouldn't see that area. Got it

PR #53, opened 2020-04-14
Feature 19 modify profile yevchuk

r-yevchuk:


server/src/main/java/com/lunchtime/service/impl/PersonServiceImpl.java
```diff
@@ -1,24 +1,53 @@
...
+        if (result.isPresent()) {
+            Person person = result.get();
+            if (personPassDto.getOldPassword().equals(person.getPassword())) {
+                person.setPassword(personPassDto.getPassword());
+                person.setName(personPassDto.getName());
+                person.setPhoneNumber(personPassDto.getPhoneNumber());
+                personRepository.save(person);
+                return personPassDto;
+            }
+            throw new Exception();
```
narekacy, line 41: Why do we need to throw here an Exception?

r-yevchuk: I throw this exception if we get wrong old password from front-end.

narekacy: May be you forgot `else if` before throwing this Exception?

r-yevchuk: This is not necessary, because we have `return`.

narekacy: Right, missed it

server/src/main/java/com/lunchtime/mapper/PersonMapper.java
```diff
@@ -0,0 +1,39 @@
...
+    public PersonDto fromPersonToDto(Person person) {
+        PersonDto personDto = new PersonDto();
+
+        personDto.setId(person.getId());
+        personDto.setName(person.getName());
+        personDto.setPhoneNumber(person.getPhoneNumber());
+        personDto.setEmail(person.getEmail());
+        personDto.setPhotoUrl(person.getPhotoUrl());
+        return personDto;
+    }
```
narekacy, line 38: I think we should change PersonMapper's logic. It shouldn't know about repository. Due to SOLID only service works with Repository. So, Mappers repsonsibility is just to map, to pass one object's value to another. Thats why if we need to map PersonDto to already existing Person, we should pass both PersonDto and Person to `fromDtoToPerson(PersonDto perstoDto, Person person)`

server/src/main/java/com/lunchtime/service/impl/PersonServiceImpl.java
```diff
@@ -1,24 +1,53 @@
...
+    public PersonPassDto updatePassword(PersonPassDto personPassDto) throws Exception {
+        Optional<Person> result = personRepository.findById(personPassDto.getId());
+        if (result.isPresent()) {
+            Person person = result.get();
+            if (personPassDto.getOldPassword().equals(person.getPassword())) {
+                person.setPassword(personPassDto.getPassword());
+                person.setName(personPassDto.getName());
+                person.setPhoneNumber(personPassDto.getPhoneNumber());
+                personRepository.save(person);
+                return personPassDto;
```
narekacy, line 39: I think PersonService shouldn't have a logic for mapping. It just has to work with repository. 
Something like this:
```
        Person oldPerson = personRepository.findPersonById(personPasswordDto.getId());
        if (personPasswordDto.getOldPassword().equals(oldPerson.getPassword())) {
            Person newPerson = personPasswordMapper.fromPasswordDtoToPerson(personPasswordDto, oldPerson);
            personRepository.save(newPerson);
            return personPassDto;
        }
```
But we need also PersonPasswordMapper here. Yes, and write not `pass` but `password` so that there won't be any misunderstanding when another developer may look your code.

narekacy:
Overall your code is good. Nevertheless, we need to review again our requirements and backend code logic. I think what I mentioned should be fixed and I'm interested in our team's opinion about this. Also I think that we don't need 'PersonPassDro' because of existing 'PersonDto'. We just can set frontend DTO with **[@JsonView](https://mkyong.com/java/jackson-jsonview-examples/)** preferences for password fields.

Also this micro-minor remarks:
- PersonController, line 46: you can use just `PersonPassDto result`, without `null`
- PersonDto, line 23: @Column here is not needed

I'm going to approve this pull request and I want our team to discuss this possible fixes soon.

narekacy (dismissed):

server/src/main/java/com/lunchtime/service/impl/PersonServiceImpl.java
narekacy: Also, if you can deal with this bug, it will be great. When I type new username, my avatar color is switching to another colors with each input symbol.

Andrii35 (dismissed):

narekacy (dismissed):

Melnyk9506:
client does not run

Melnyk9506:
Failed to compile.

./src/components/profile/pass-change.js
Module not found: Can't resolve 'react-password-strength-bar'

r-yevchuk:
> Failed to compile.
> 
> ./src/components/profile/pass-change.js
> Module not found: Can't resolve 'react-password-strength-bar'

npm install

narekacy (approved):

Andrii35 (approved):

Melnyk9506 (approved):

PR #54, opened 2020-04-15
Fix 22. Fix leave feedback to restaurant as an authorized user. narek

narekacy:
Non authorized user can **only see** restaurant feedback. 
Authorized user can **see and leave** feedback for restaurant.


Andrii35 (commented):
I think, it will be better, when Events font will be yellow such on other headers in our app.

narekacy:
> I think, it will be better, when Events font will be yellow such on other headers in our app.

Very good idea! Will fix it. Thanks!

VozNick (dismissed):

Andrii35 (approved):

Melnyk9506 (approved):

VozNick (approved):

PR #65, opened 2020-04-21
Fix issue hardcode user

Melnyk9506:
guys this task is done for profile and sending feedback


narekacy:
Overall evetything is working, I approve this PR.

After demo, we should care if we really need this `getPersonDtoByEmail()` method. We can 'get' in Controller and 'find' in Service. Moreover, we try to find not 'dto' in DB but 'entity'. Dto is just mapper's return value for frontend. 

Also I'm not sure we can make method below public as it is package-private n `JwtUtils` as default . If no, we may create another service class for this purpose and extend this method there.
```
public String extractEmail(String token) {
        return extractClaim(token, Claims::getSubject);
    }
```

Keep in mind to name branch due to task name.

narekacy (dismissed):

VozNick (dismissed):

VozNick (approved):

narekacy (approved):

Andrii35 (approved):

PR #70, opened 2020-04-22
Fix 126. Modify date and time view in restaurant feedback. narek

narekacy:
Just showed date first and then time. This view would be better.

### DIFF
--- a/client/src/components/feedback/comment.js
+++ b/client/src/components/feedback/comment.js
@@ -14,7 +14,7 @@ class FeedbackComment extends Component {
           <small className="date">
             {
               // eslint-disable-next-line max-len
-            `${item.date.substring(11, 16)} ${item.date.substring(8, 10)}/${item.date.substring(5, 7)}/${item.date.substring(0, 4)}`
+            `${item.date.substring(8, 10)}/${item.date.substring(5, 7)}/${item.date.substring(0, 4)} ${item.date.substring(11, 16)}`
           }
           </small>
           <hr className="hrStyle" />
--- a/client/src/components/feedback/comment.js
+++ b/client/src/components/feedback/comment.js
@@ -13,8 +13,8 @@ class FeedbackComment extends Component {
           <strong className="name">{item.person.name}</strong>
           <small className="date">
             {
-              // eslint-disable-next-line max-len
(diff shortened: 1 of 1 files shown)


client/src/components/feedback/comment.js
```diff
@@ -14,7 +14,7 @@ class FeedbackComment extends Component {
           <small className="date">
             {
               // eslint-disable-next-line max-len
```
vitaliimak: You can use `+` operator to concatenate two template strings

client/src/components/feedback/comment.js
```diff
@@ -14,7 +14,7 @@ class FeedbackComment extends Component {
...
-            `${item.date.substring(11, 16)} ${item.date.substring(8, 10)}/${item.date.substring(5, 7)}/${item.date.substring(0, 4)}`
+            `${item.date.substring(8, 10)}/${item.date.substring(5, 7)}/${item.date.substring(0, 4)} ${item.date.substring(11, 16)}`
```
vitaliimak: You can create Date object from `item.date` and create date view from it
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat
Using `substring()` isn't good decision in this case

vitaliimak (dismissed):

client/src/components/feedback/comment.js
narekacy: Due to tradition, when I try this - it doesn't work. But when you give an advice - it works! Thanks!

client/src/components/feedback/comment.js
narekacy: With **[these](https://www.w3schools.com/js/js_date_methods.asp)** methods code would be much more.

vitaliimak: Maybe, but in that case, you will work with date object and won't use unpredictable strings `2020-04-23T06:58:19.663Z` or `2020-04-23T6:58:19.663Z`

And it will be simpler to update date view. Now you use magic constants `substring(11, 16)` 11? 16?

vitaliimak (approved):

VozNick (approved):

r-yevchuk (approved):

PR #71, opened 2020-04-23
Fixes 130 132 yevchuk

r-yevchuk:


vitaliimak (changes requested):
Could you divide this PR to 2 independent PR? As you have 2 tasks in trello

r-yevchuk:
> Could you divide this PR to 2 independent PR? As you have 2 tasks in trello

At first, I disagreed. But later I changed my mind. I think you are right. One task is one branch

PR #72, opened 2020-04-23
Fix hover button style

r-yevchuk:


client/src/styles/app.css
```diff
@@ -24,10 +24,15 @@ body{
 }
 
 .btn-restaurants{
-  font-size: 150%;
+  font-size: 160%;
 }
 
 .btn-restaurants:hover{
```
vitaliimak, line 31: image

Button is almost invisible over this background

vitaliimak (changes requested):

vitaliimak (approved):
Cool! I can see text) 
But, you use very different styles on similar buttons, it isn't very good

There are github buttons
image
image
image
image
A lot of pages, but button the same

Melnyk9506 (approved):
that ok

VozNick (approved):

PR #73, opened 2020-04-23
Feature 133 fix token time limit

Melnyk9506:
guys max token time limit 24 days  for now

### DIFF
--- a/client/src/services/person.js
+++ b/client/src/services/person.js
@@ -5,7 +5,6 @@ import Auth from './auth';
 class Person {
   constructor() {
     this.userInfo = {};
-    this.userKey = 'user';
   }
 
   getProfile() {
@@ -18,7 +17,6 @@ class Person {
             return;
           }
           this.userInfo = response.data;
-          localStorage.setItem(this.userKey, response.data);
         });
     }
   }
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/lunchtime/security/JwtUtil.java
```diff
@@ -21,7 +21,9 @@
     private int miliseconds = 1000;
     private int seconds = 60;
     private int minutes = 60;
-    private int hours = 10;
+    private int hours = 24;
+    private int days = 24;
+    private long timeLimit = miliseconds * seconds * minutes * hours * days;
```
vitaliimak: Why you use both `long and int`?
Maybe it will be better to create privat method `getTimeLimit()`?

client/src/services/person.js
```diff
@@ -18,7 +17,6 @@ class Person {
             return;
           }
           this.userInfo = response.data;
-          localStorage.setItem(this.userKey, response.data);
```
vitaliimak, line 21: Why you remove this functionality in this PR? You should create new task for it

vitaliimak (changes requested):
Don't fix a lot of issues in one PR

vitaliimak (approved):

r-yevchuk (approved):

VozNick (approved):

PR #79, opened 2020-04-24
Fix 131 remove additional login button makarchenkov

vitaliimak:


narekacy (commented):
I think it is possible to use `onClick()` function here. This will help us to maintain **[these](https://www.w3schools.com/js/js_htmldom_events.asp)** methods to make better UI.

It will be better to inform user that he should log in to be able to make order.

Also my IDE hints in line 59 this: `Element Tabs doesn't have required attribute id `

vitaliimak:
> I think it is possible to use onClick() function here. This will help us to maintain these methods to make better UI.

We use react-router-dom

> It will be better to inform user that he should log in to be able to make order.

My task is making button disabled, when user is unauthorized. Please, read task description

> Also my IDE hints in line 59 this: Element Tabs doesn't have required attribute id 

You can create new task in Trello

narekacy:
> We use react-router-dom

It doesn't work with `onClick()`?

> My task is making button disabled, when user is unauthorized. Please, read task description

Anyway, it's very easy to add feature with informing but this will make UX much better.

> Also my IDE hints in line 59 this: Element Tabs doesn't have required attribute id

Do your IDE shows some hints there?

vitaliimak:
> It doesn't work with `onClick()`?

Why I should use `onClick` if Link works well?

> Anyway, it's very easy to add feature with informing but this will make UX much better.

Yes, I'm agree, but we decided not to do it (task description) on planing meeting

> Do your IDE shows some hints there?

Nope, and it isn't my task

narekacy:
IMO, message for user must be. If you agree, so another task in trello should be already there.

I can't remind myself how we missed this on planning day. So our team should discuss it again later.

r-yevchuk:
I also think you need to inform the user
using some kind of Alert or Badge

vitaliimak:
@r-yevchuk On planing meeting I suggested few approaches (make button disabled and show alerts), @narekacy suggested that that alert can be showed when user move cursor to disabled button (hover style)
But we all together have decided to use only disabled buttons, why your opinion is more important than team decision?

vitaliimak:
Added alert for unauthorized users

r-yevchuk (approved):
Cool!

VozNick (approved):

Melnyk9506 (approved):

PR #82, opened 2020-04-26
Feature 138. Set 'USER' role for registered user. narek

narekacy:


server/src/main/java/com/lunchtime/models/Person.java
```diff
@@ -69,13 +69,6 @@
     @Column(name = "status", columnDefinition = "varchar(50) default 'ACTIVE'")
     private Status status;
 
-    @ManyToOne(fetch = FetchType.EAGER)
-    @JoinTable(name = "user_roles",
```
VozNick, line 73: You deleted user_roles join table from model class. But in database (dev) this table will still be there. I think without script (to remove this table in db) it will make error.

r-yevchuk: I also think you need to delete the unused table.

narekacy: It doesn't make error on my environment. I agree, we should remove it from scripts.

Thanks, fixed this yet. Now you can check again.

r-yevchuk (approved):

VozNick (approved):

vitaliimak (approved):

PR #83, opened 2020-04-27
Feature 84 change avatar yevchuk

r-yevchuk:


vitaliimak (changes requested):
image

vitaliimak:
Sorry, I thought that your task is ready to review, I didn't notice that it placed in "In progress"

r-yevchuk:
Fixed

server/src/main/java/com/lunchtime/config/ResourcesPath.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.springframework.context.annotation.Configuration;
+
+import java.io.File;
+
+@Configuration
+public class ResourcesPath {
+
+    public static String getResourcePath() {
+        return new File(System.getProperty("user.dir")).getParent() + "/static/";
```
vitaliimak: You shouldn't use `getParent()` method, because application can be run in root folder (`/`)
image
image
Or make your code more flexible

In our staging server application runs in folder `/app` by `appuser` and doesn't have rights to create `/static` folder. You also should update docker-compose file - add new `static` volume, which must be mapped to real filesystem
`
volumes:
      - ./server/static:/app/static
`
Without it, all users files (avatars) will be removed after next update

vitaliimak (changes requested):
image

Is this nice cat need? static/images/profile/5.jpg

image
folder static should be in gitignore

image
it doesn't work in docker environment

You shouldn't commit `static` folder and all images to git repository. We saved images in `resources` folder, because we didn't know other path and we used seed file for creating testing data. But for real data (like clients photos) we need to use another path. We should save images only in servers, but not in repository

client/src/components/profile/profile.js
```diff
@@ -54,20 +54,46 @@ class Profile extends Component {
           user: response.data,
           isFetching: true,
         });
+        this.getAvatarUrl(response.data);
       });
   }
 
-  saveAlertState(show) {
+  getAvatarUrl(user) {
+    if (user.photoUrl === undefined || user.photoUrl === null) {
```
vitaliimak, line 62: Simpler and safer
`if (user && user.photoUrl && user.photoUrl.length)`

It is JavaScript user can be `null` and if `photoUrl` is string, it must have lenngth > 0

server/src/test/java/com/lunchtime/stub/PersonServiceStub.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.lunchtime.models.Person;
+import com.lunchtime.service.PersonService;
+import com.lunchtime.service.dto.PersonDto;
+import com.lunchtime.service.dto.PersonPassDto;
+import com.lunchtime.service.dto.RegisterPerson;
+
+import java.util.Optional;
+
+public class PersonServiceStub implements PersonService {
```
vitaliimak: Why you create this class, but don't use it?

client/src/components/profile/profile.js
vitaliimak: Sorry, wrong file. This comment must be here https://github.com/Social-projects-Rivne/Lunch_Time/pull/83/files#diff-eeea51dd0a4695691318601103aeefe8R20

r-yevchuk: Fixed

server/src/test/java/com/lunchtime/stub/PersonServiceStub.java
r-yevchuk: File deleted

vitaliimak (approved):

VozNick (approved):

Melnyk9506 (approved):
Well done good job!

PR #76, opened 2020-04-23
Fix 103 menu view nazaruk

Andrii35:
Add drop-down
Fix view (background without image)
Don't need here DB migration 
Price validation, pagination, English gramms and images saved local are allready implemented.

PR #88, opened 2020-04-29
Feature 137. Add menu dishes to order. narek

narekacy:
This task is done due to requirements. Yes, there are a lot of things to improve. We will consider it later. 
Tests are not done as I'm not almostly responsible for them right now. We can do test task in another one or like bug fix task.

Json type is used for persisting ordered dishes in 'Order' entity. This wouldn't be easier with Map or another table. In future, to retrieve data from DB entity property we can use special way to **[implement](https://stackoverflow.com/a/24012023/13334444)** it. Also I thought whether migration naming is used correctly and I **[think](https://flywaydb.org/documentation/migrations#naming-1)** that it is.

We should implement another tasks for:
- when user isn't logged in, 'Add' button on menu item dish should be disabled with some notification as **[here](https://github.com/Social-projects-Rivne/Lunch_Time/pull/79/commits/fdee0566bf7acfc015e9fb07b5e9cc37ba26bbee)**;
- 'Complete order' button should be on the right side. Have no idea why my `marginRight` doesn't work **[here](https://github.com/Social-projects-Rivne/Lunch_Time/blob/baafa48fcffb06e9668a7ecacd30e732419fd2c3/client/src/components/menu-views/menu-view.js#L128)**;
- when item is choosed multiple times (e.g. 'Pizza Margarita 3 times' = quantity 3), we should show it like 'Pizza Margarita x3', but not as 'Pizza Margarita, Pizza Margarita, Pizza Margarita';
- user notification when order is done successfully;
- possibility to change dishes list or quantity;
- person id in local storage. If you are just logged in user, when you create order - your id is sent to server inside 'Order'. But when you reload page, your id is undefined. So, we need to improve user's id persisting in local storage to prevent such behaviour when app is reloaded.

However, most of this should be considered to UX before planning tasks.

**EDITED 02.05.2020 10:33 PM:**
When user creates order, he receives response with all content of order. It means that he receives also all order dishes with inside order and inside person values, so repsonse entity is too big. This can be fixed when Order field in OrderDish entity is switched to Long orderId. I will kindly appreciate reviewers opinion for this.

**EDITED 04.05.2020 10:39 PM:**
Tests are fixed so that they can be passed right now. I'm sure they're not working as needed. For now we can see that another functionality works ok and create another task for further tests fixing if needed.


client/src/components/menu-views/menu-view.js
```diff
@@ -58,24 +65,76 @@ class Menu extends Component {
...
+              menuItemDishesMap: menuItemDishesMap,
+            },
+          }}
+          >
+            {this.state.dishes.length > 0 && (
+              <Button
+                className="complete"
+                variant="primary"
+                style={{
+                  marginRight: 40,
```
vitaliimak: One of possible decision
```
<div style={{
            display: 'flex',
            justifyContent: 'flex-end',
          }}
          >
            <Link to={{
              pathname: `/restaurants/${id}/new-order`,
              state: {
                restaurantName: name,
                dishes: dishes,
                menuItemDishesMap: menuItemDishesMap,
              },
            }}
            >
              {this.state.dishes.length > 0 && (
                <Button
                  className="complete"
                  variant="primary"
                >
                  Complete order (
                  {this.state.dishes.length}
                  {' '}
                  items)
                </Button>
              )}
            </Link>
          </div>
```

vitaliimak (commented):
> when user isn't logged in, 'Add' button on menu item dish should be disabled with some notification as here;

I think this PR should be merged after this one https://github.com/Social-projects-Rivne/Lunch_Time/pull/76 In PR #76 buttons `Add` set as disabled for unauthorized users. In this PR unauthorized use can open new Order page

Register new person and already tried to create order
image
The same error after logout and login

client/src/components/menu-views/menu-view.js
narekacy: Thanks, fixed!
Also fixed register user bug. But, the issue with refreshing app still exists.

server/src/main/java/com/lunchtime/service/impl/OrderServiceImpl.java
```diff
@@ -26,42 +29,49 @@
...
             return null;
         }
-        if (order.getStartTime().after(order.getFinishTime())) {
+        if (orderDto.getStartTime().after(orderDto.getFinishTime())) {
             return null;
         }
 
         List<Order> orders = orderRepository
-            .findAllOrdersByTableInTime(order.getTable().getId(), order.getStartTime(), order.getFinishTime());
+            .findAllOrdersByTableInTime((long) orderDto.getTableId(), orderDto.getStartTime(), orderDto.getFinishTime());
```
vitaliimak: Useless type casting, method `getTabletId()` returns Long

client/src/components/order/new-order.js
```diff
@@ -234,6 +251,16 @@ class NewOrder extends Component {
             />
           </Form.Group>
 
+          <Form.Group>
```
vitaliimak, line 254: Don't show it input, if there aren't dishes. Or make it interactive (when user click on this input add dish)
image

client/src/components/order/new-order.js
```diff
@@ -73,13 +75,28 @@ class NewOrder extends Component {
     } else {
       return;
     }
+    let orderedDishes = this.props.location.state.menuItemDishesMap;
```
vitaliimak, line 78: Simpler and more readable [{dishId: 25, quantity: 3}, {dishId: 26, quantity: 2}, {dishId: 29, quantity: 2}]
image

server/src/main/java/com/lunchtime/models/Order.java
```diff
@@ -47,6 +48,13 @@
     @JoinColumn(name = "status_id", referencedColumnName = "id")
     private OrderStatus status;
 
+    @OneToMany(
+        mappedBy = "order",
+        cascade = CascadeType.ALL,
+        orphanRemoval = true
+    )
+    private List<OrderDish> orderDishList = new ArrayList<>(); //?
```
vitaliimak: Could you remove this comment?

vitaliimak (changes requested):
image
Tests fail
image
Don't add images from static/images/ folder to repository
image

client/src/components/order/new-order.js
narekacy: It isn't diffucult to hide this, but as we want to show user that he can add dishes to his order, this input should be visible. I think we will add similar functionality in our app.

client/src/components/order/new-order.js
narekacy: This json is automatically converts from map object. To implement another view we should add some logic to save our Map in List both frontend and backend. This isn't necessary.

server/src/main/java/com/lunchtime/models/Order.java
narekacy: Sorry, forgot about this.

client/src/components/order/new-order.js
vitaliimak: For now it looks like input, but it is disabled. If you want to show it all time, you should add some description (message, alert) to explain what it is. Or you can add redirect to dishes page from this form
But you shouldn't show blank, disabled and useless input on page

client/src/components/order/new-order.js
vitaliimak: In this case my comment is only recommendation, you don't need to follow it

r-yevchuk (approved):
I approve it, but I also don't like what the "ordered dishes" input looks like

VozNick (approved):

vitaliimak (approved):

PR #91, opened 2020-04-29
Fix 149. Add new regex for password. narek 

narekacy:
Now if user wants to register with password consisting whitespace as first/last symbols, he can't do that. This check is done when user 'Submits' registration form and on the backend


vitaliimak:
@narekacy could you merge current dev branch into your feature branch?

vitaliimak (commented):
Thanks, done yet!

VozNick (dismissed):

narekacy:
My password has more than 8 symbols and strong enough  

image

For better UX positive messages will be better, ("You password should be stronger and contain more than 8 symbols")
Password validation in back-end works well

Answering @vitalii-mak, I've changed this user message and it shows new one now.

I wanted to 'quote answer', but edited previous comment, afterwards wanted to delete my edition, but in fact deleted first comment. Fortunately, our git channel keeps comments history, so I attached this here again.

client/src/components/register-login/register.js
```diff
@@ -52,7 +52,7 @@ class Register extends Component {
       showStrong: false,
       passwordStrength: '',
       passwordInputStarted: false,
-      passwordInputTitle: "Password shouldn't be weak and less than 8 symbols",
+      passwordInputTitle: 'We recommend setting strong password with 8 or more symbols',
```
vitaliimak: You should explain user when he or she wrongs. This message still doesn't have any information about start or ending with spaces. You can use separate message in that specific case

vitaliimak: I have 10
image

narekacy: Done

vitaliimak (approved):

VozNick (approved):

r-yevchuk (approved):

PR #92, opened 2020-04-29
Feature 36 owners restaurant list voznyarskyy

VozNick:


vitaliimak:
@VozNick your task in "In progress"

vitaliimak:
@VozNick could you merge current dev brunch into your feature branch?

And I've updated role, but databse doesn't change
image
image

client/src/components/profile/profile.js
```diff
@@ -26,14 +27,30 @@ class Profile extends Component {
       user: {},
       isFetching: false,
       isShowAlert: false,
+      isOwnerRole: 2, // id that refers to Restaurant Owner role in DB
```
vitaliimak: You can create separate file with constants where initialize this general constant and use better name (**is**Something - means true or false)
As you use id in DB, it can be changed in different environments (e.g. in other database, like MySql, in production or testing environments)

client/src/components/restaurant/restaurant-item.js
```diff
@@ -36,14 +37,25 @@ class Restaurant extends Component {
         this.setState({
           restaurant: response.data,
           isFetching: true,
+          // eslint-disable-next-line eqeqeq
```
vitaliimak: Nice! Very interesting message)
You should use `===` operator and explicit type conversion (e.g. `+localStorage.getItem('userID')`

client/src/components/restaurant/restaurant-item.js
```diff
@@ -36,14 +37,25 @@ class Restaurant extends Component {
...
   render() {
-    const { isFetching, restaurant } = this.state;
+    const { isFetching, restaurant, isOwner } = this.state;
     const { match: { params: { id } }, isAuthenticated } = this.props;
 
+    let isOwnerText;
+    if (isOwner) {
+      isOwnerText = (
+        <Container>
+          <h4>(You are the owner of this restaurant)</h4>
```
vitaliimak: Why you use brackets?

client/src/services/auth.js
```diff
@@ -29,6 +29,7 @@ class Auth {
   removeToken() {
     this.token = '';
     localStorage.removeItem(this.tokenKey);
+    localStorage.removeItem('userID');
```
vitaliimak: If you update this method, you should change there name. It remove not only token now

vitaliimak:
Sorry, I thought that your task is ready to review. I late noticed that it is still in "In progress"

client/src/components/restaurant/restaurant-item.js
VozNick: They will be shown on rest page. Without them the taxt may look like title or smth else (part of our UI design)...
This text must be like a note. Insted brackets we can put "!" or "*" in front of text, to show thats it's just a note for owner,

client/src/services/auth.js
VozNick: Not understood) 
when user logs in, local storage saves token and userId. Then user logs out, and local deletes token, and it needs to delete id..

vitaliimak: You can rename method removeToken -> clearStorage

client/src/components/restaurant/restaurant-item.js
vitaliimak: You can use specific style, other background color, or something else

server/src/main/java/com/lunchtime/controllers/RestaurantController.java
```diff
@@ -66,6 +66,16 @@ public RestaurantController(RestaurantService restaurantService) {
...
+    @GetMapping("/userId")
+    public ResponseEntity<Page<Restaurant>> getRestaurantPageByUserId(
+        Long userId, Pageable pageable) {
+        Page<Restaurant> restaurantPage = restaurantService.getRestaurantPageByUserId(userId, pageable);
+        if (restaurantPage == null) {
+            return ResponseEntity.notFound().build();
+        }
+        return ResponseEntity.ok().body(restaurantPage);
+    }
+
```
narekacy, line 78: I think this method's name should be 'getRestaurant**s**ByUserId', without 'Page', as well as getRestaurant**s**ByUserId'

server/src/main/java/com/lunchtime/mapper/PersonMapper.java
```diff
@@ -26,6 +26,7 @@ public Person fromDtoToPerson(PersonDto personDto) {
             person.setPhoneNumber(personDto.getPhoneNumber());
             person.setEmail(personDto.getEmail());
             person.setPhotoUrl(personDto.getPhotoUrl());
+            person.setRoleId(personDto.getRole());
             return person;
```
narekacy, line 30: Nothing serious, but we should name setters exactly what we mean. I'd prefer `roleId` field, as I did in my task

image

server/src/main/java/com/lunchtime/repository/RestaurantRepository.java
```diff
@@ -13,4 +14,17 @@
...
+     * It also returns all records that are NOT marked as deleted.
+     *
+     * @param id       user (owner) whose restaurants needs to be return
+     * @param pageable identifies what page needs to be returned and with what size
+     * @return the restaurant list of concrete user (owner)
+     */
+    @Query("select r from Restaurant r "
+        + "where r.personId in :id "
+        + "and r.isDeleted = false")
+    Page<Restaurant> findByOwnerRestList(Long id, Pageable pageable);
```
narekacy, line 29: I think `findRestaurantsByUserId` method name would be better. Our 'owner' is user at all.

server/src/main/java/com/lunchtime/repository/PersonRepository.java
```diff
@@ -11,4 +12,9 @@
     Person findFirstByEmail(String email);
 
     Person findPersonByPhoneNumber(String email);
+
+    @Query("select p.roleId from Person p "
+        + "where p.id in :userId "
+        + "and p.isDeleted = false")
+    long findByRoleId(long userId);
```
narekacy: If we want to find role id by user id, this method's name should have `findRoleIdByUserId` name. Thus, this method should be in RoleRepository as we want to find Role, but nor Person (... due to `select p.roleId` in query).

server/src/main/java/com/lunchtime/service/impl/RestaurantServiceImpl.java
```diff
@@ -13,10 +14,16 @@
...
 
-    public RestaurantServiceImpl(RestaurantRepository restaurantRepository, PersonService personService) {
+    public RestaurantServiceImpl(
+        RestaurantRepository restaurantRepository,
+        PersonRepository personRepository,
+        PersonService personService) {
         this.restaurantRepository = restaurantRepository;
+        this.personRepository = personRepository;
         this.personService = personService;
     }
```
narekacy: **[@RequiredArgsConstructor](https://projectlombok.org/features/constructor)** will help us to make code more readable. We need just `@RequiredArgsConstructor` annotation and no args constructor here, as `@RequiredArgsConstructor` automatically generates constructor with `final` fields.

server/src/main/java/com/lunchtime/service/impl/RestaurantServiceImpl.java
```diff
@@ -33,6 +40,14 @@ public Restaurant saveRestaurant(Restaurant restaurant) {
         return restaurantRepository.findByIsDeletedFalse(pageable);
     }
 
+    public Page<Restaurant> getRestaurantPageByUserId(Long userId, Pageable pageable) {
+        long userRoleId = personRepository.findByRoleId(userId);
+        if (userRoleId == OWNER_ROLE_ID) {
+            return restaurantRepository.findByOwnerRestList(userId, pageable);
+        }
+        return null;
+    }
```
narekacy, line 42: I'm not sure if we should store `OWNER_ROLE_ID` here. We can get this value from frontend constants right now. But, honestly, I'd get thid id value from Java Object Enums that can be stored in DB with its persisting id. So, we can check users role withoud access to DB in line 44. Haven't implemented smth like that, but I think this is possible.

client/src/app.js
```diff
@@ -40,6 +40,7 @@ class App extends Component {
     this.setState({ isAuthenticated: false }, () => {
       Auth.removeToken();
     });
+    localStorage.removeItem('userID');
```
vitaliimak: Very bad decision, your main class is responsible to remove information from localstorage. If we will change storage system, we 100% will forget to fix this line.
Your previous approach was well enough. You just needed to change method name

client/src/components/profile/profile.js
```diff
@@ -18,6 +18,7 @@ import PhotoEditor from './avatar';
 import InfoChange from './info-change';
 import Auth from '../../services/auth';
 import OwnersRestaurants from './owners-restaurants';
+import { OWNER_ROLE_ID } from '../../constants';
```
vitaliimak, line 21: Cool!

client/src/components/profile/owners-restaurants.js
```diff
@@ -0,0 +1,118 @@
+import React, { Component } from 'react';
```
vitaliimak, line 1: Could you remove background image from this page?
image

vitaliimak: Don't show pages if restaurant list is empty
image

server/src/main/java/com/lunchtime/repository/PersonRepository.java
vitaliimak: @VozNick you should change method name or signature to make it more clear to other developers

server/src/main/java/com/lunchtime/repository/RestaurantRepository.java
vitaliimak: @narekacy We don't have `users` we have 'persons'

narekacy: It doesn't matter, let us use 'persons'

server/src/main/java/com/lunchtime/controllers/RestaurantController.java
VozNick: without page, you won't no how many rests will be returned (all of them or some size). 
Second, this method returns not only list of rests, but page (it has many other fields for page like size, count, number, sort ,ets)

server/src/main/java/com/lunchtime/mapper/PersonMapper.java
VozNick: you're right, fixed)

server/src/main/java/com/lunchtime/repository/PersonRepository.java
VozNick: it won't work in RoleRepository. It 's not connected to user. I'll change the name

server/src/main/java/com/lunchtime/repository/RestaurantRepository.java
VozNick: person is general name for owner, waiter, user (can be general in some way)), ets..
"findByUser" ... it's unclear what this method returns: a list of the owner or the waiter who serves the specific rest list

server/src/main/java/com/lunchtime/service/impl/RestaurantServiceImpl.java
VozNick: maybe we could get it from frontend, but in case for example in postman I can put some other number, and it wouldn'd work
Earlier we already had CategoryEnum for event, and we moved this to DB. This is a same situation)

client/src/services/auth.js
VozNick: @vitalii-mak if insted this line, i'll put localStorage.clear();
i think there would be no need to rename. yes? if i rename to clearStorage, it's quite not understandable I think

client/src/services/auth.js
```diff
@@ -29,6 +29,7 @@ class Auth {
   removeToken() {
     this.token = '';
     localStorage.removeItem(this.tokenKey);
+    localStorage.clear();
```
vitaliimak: Please, please, please! Change method name!

client/src/components/profile/owners-restaurants.js
```diff
@@ -78,6 +78,14 @@ class OwnersRestaurants extends Component {
     );
   }
 
+  initTextNoRestaurants() {
+    return (
+      <Alert variant="info">
+        You have no restaurants yet !
```
vitaliimak: Could you remove white space before exclamation mark? In both alerts

vitaliimak (changes requested):

vitaliimak (approved):

r-yevchuk (approved):

Melnyk9506 (approved):

PR #96, opened 2020-04-30
Fix 115 menu view nazaruk

Andrii35:


r-yevchuk:
I get this error 
Знімок екрану_2020-05-01_13-10-26

Andrii35:
> I get this error
I have fixed it. This problem is not on staging server only on dev.

r-yevchuk (approved):

server/src/main/java/com/lunchtime/models/CategoryFood.java
```diff
@@ -3,7 +3,6 @@
 import javax.persistence.*;
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.Size;
-
```
vitaliimak, line 6: IDEA has very useful keybindings `Ctrl+Alt+O` and `Ctrl+Alt+L`
This line divided imports from different packages

vitaliimak (approved):

VozNick (approved):

PR #104, opened 2020-05-06
fixed dropdown, image font and align image

Andrii35:


vitaliimak (dismissed):
It would be better to make columns #2 and #3 wider and columns #4, 5 and 6 narrower. As last columns don't have a lot of text inside them
image

Andrii35:
> It would be better to make columns #2 and #3 wider and columns #4, 5 and 6 narrower. As last columns don't have a lot of text inside them
> image

I have made columns #2 and #3 widder, i think it looks better.

narekacy (approved):

vitaliimak (approved):
It doesn't look well on small screens :(
image

Melnyk9506 (approved):

PR #110, opened 2020-05-07
Feature 170 events search yevchuk

r-yevchuk:


narekacy:
This is interesting task solution, as we fetch needed event from already fetched events. Something like caching :)

However, our events fetching type should be lazy with pageable option. This will show searched events from users current page. Even if we have eager fetch type (what is not good in events case), we can see all events when searching. But, we can't see new events after our page is loaded. 

I mean, if user left his device for 10 minutes, and new events are uploaded (or removed) during this 10 minutes, user can't access relevant events in search.

Also, once page is loaded, I turn off running server. And events search still works, but not all events title images are loaded from server.

Summarizing, this task meets its requirements, especially that pageable events request may be not implemented. Anyway, I think search should work together with server part.

It costs me nothing to approve this pull request right now, but let us see what other developers think about this.

Andrii35 (approved):
I'm agree with Narek. The search must interact with the server. I can find event by searching. Task has't detail description. So I will Implement it now and comments can be implemented by fix.

narekacy (approved):

Melnyk9506 (approved):

PR #111, opened 2020-05-08
Add autohide toast

vitaliimak:
It isn't best solution, but this toast appears on the right side and doesn't push the rest part of page a lot.
image

### DIFF
--- a/client/src/components/feedback/comment.js
+++ b/client/src/components/feedback/comment.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import {
-  Container, Image, Badge,
+  Container, Image, Badge, Toast,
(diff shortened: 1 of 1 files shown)


narekacy:
Actually, I wonder why this branch doesn't show correctly code updatind. Anyway, I saw what Toast does. Not too much, but it pushes the rest part of content. I tried to fix this with styles, but nothing happened.

To be honest, I'd prefer micro memory leak in return for better UI. I agree that timeouts are really bad for application even if this is not serious issue right now. I really appreciate your opinion and new branch to show better solution, but now I think we should also care about UI

vitaliimak:
@narekacy Ok

PR #116, opened 2020-05-10
Fix localstorage get

vitaliimak:
Owner doesn't see alert with information about his or her restaurants

### DIFF
--- a/client/src/components/feedback/send.js
+++ b/client/src/components/feedback/send.js
@@ -10,7 +10,6 @@ import Auth from '../../services/auth';
 class FeedbackSend extends Component {
   constructor(props) {
     super(props);
-    this.userID = null;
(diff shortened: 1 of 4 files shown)


narekacy (approved):

narekacy:
It is ok, very good that you noticed this bug.

I haven't found this task in trello. Could you create it if it doesn't exist?

Melnyk9506 (approved):

vitaliimak:
@narekacy no, I don't have time, you can do it, if you want

narekacy:
> @narekacy no, I don't have time, you can do it, if you want

I think we need it in trello anymore. Ok, I will do that.

Andrii35 (approved):

